PR #9, opened 2024-09-25
Mpi for HESS/NAC

karmachoi:
Jingbai added MPI options for Hessian and NAC calculations.

- Both Hessian and NAC (Nonadiabatic Coupling) utilizes MPI for optimal parallel efficiency.


karmachoi:
Please revert the delete directories of NAC and HESS of example directory.

karmachoi:
All test files are passed without any errors.

PR #38, opened 2025-01-16
Ecp

mohsenkor:


karmachoi:
Please add descriptions.

karmachoi:
Also it would be useful to printout the detailed basis set info in log file right after basis set reading.

Issue #37, opened 2024-12-19
charges in openqp


stecrespi:
Possibly a silly question...I am running the examples of H2O in the mrsf folder and I noticed that the energy job produced a detailed summary of the single point, including charges, while the gradient (and the optimization) skips that part completely.
I checked the variables that are set as default for all the runs and in all cases properties.scf_prop=el_mom,mulliken
I tried to manually set in the input:

[properties]
scf_prop=mulliken

but nothing changes in my output (so, no charges after the optimization or gradient)
Am I doing something wrong? 
I also tried combinations such as

[properties]
export=True
td_prop=True
scf_prop=mulliken

but no great success...

Thank you for the help!!
Stefano

karmachoi:
Does this mean that you don't get charges after gradient/geometry optimization? Ali, please check the python layer.

stecrespi:
That is indeed the case!

Alireza-Lashkaripour:
Dear @stecrespi 
The changes are implemented, you can ''git pull" and get the properties printed in gradient/opt log file.

PR #42, opened 2025-01-22
Now the Prop. will be printed after gradient and Opt calcualtion

Alireza-Lashkaripour:
https://github.com/Open-Quantum-Platform/openqp/issues/37 is solved.

### DIFF
--- a/pyoqp/oqp/library/runfunc.py
+++ b/pyoqp/oqp/library/runfunc.py
@@ -56,6 +56,9 @@ def compute_grad(mol):
     # compute gradient
     Gradient(mol).gradient()
 
+    # compute properties
+    compute_scf_prop(mol)
+
     # compute dftd4
     LastStep(mol).compute(mol, grad_list=mol.config['properties']['grad'])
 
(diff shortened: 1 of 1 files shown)


Alireza-Lashkaripour:
Please don't mind this part, I will update docker after merge. 
image

Alireza-Lashkaripour:
After discussion with @mohsenkor and checking the code, apparently we don't have properties function for response part (td_prop), I has been mentioned in the Wiki too. 

```
- **scf_prop**: Compute properties of the reference state.
  - **Options**:
    - `el_mom`: Electronic moments.
    - `mulliken`: Mulliken charges.
  - **Default**: `scf_prop=el_mom,Mulliken` (computes both)
- **td_prop**: Compute properties of the response state.
  - **Options**:
    - `True`: Compute electronic properties for the response state. _(Not available yet)_
    - `False`: Do not compute electronic properties for the response state.
```
    

So we can merge this changes for printing scf_prop and try to implement the td_prop.

PR #51, opened 2025-02-13
Add initial CI working on GitHub runners

foxtran:
No checks for tests verification


Alireza-Lashkaripour:
I guess we need one more line for installing cffi.

Alireza-Lashkaripour:
Maybe changing OS runner to ubuntu-latest can solve the problem (as it works for the docker workflow):
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow

toxtran:
That is Python versions compability issue. I just need to check a proper one.

PR #54, opened 2025-02-17
fix: resolve unexpected memory increase

mohsenkor:
int2_data was not properly cleaned up and deallocated in tdhf_mrsf_z_vector.F90 and tdhf_sf_z_vector.F90.


karmachoi:
Great!

karmachoi (approved):
Great work!

PR #59, opened 2025-02-20
pFON cooling (new approach) 

Alireza-Lashkaripour:
+ some small corrections for examples.


mohsenkor:
if (occ_a(i) > 1.0e-14_dp) then 
                    do mu = 1, nbf
                        do nu = 1, nbf 
                            dtmp(mu,nu) = dtmp(mu,nu) + occ_a(i) * mo_a(mu,i)*mo_a(nu,i)
                        end do 
                    end do 
                 end if
         ==>
          if (occ_a(i) < 1.0e-14_dp) cycle

PR #61, opened 2025-02-21
add forced_attempt to restart scf

lijingbai2009:
Add forced_attmept to restart the SCF calculation if the previous is not converged. Useful for with the pFON method


Alireza-Lashkaripour (approved):
I have tested and added this into wiki.

PR #64, opened 2025-02-27
revised the guess function for single point class

lijingbai2009:
guess preparation functions (set_basis, int_1e, guess) are moved into the single point class for more flexible control of guess calculation. 
No longer need to call set_basis, int_1e, or guess before single-point calculations.
Also leave a todo mark in _init_convergence function for calling basis set projection when it is available.


Alireza-Lashkaripour (commented):
Example for using basis library in both [system] and [scf] section added.

PR #65, opened 2025-02-27
wb97xd family are added 

Alireza-Lashkaripour:


source/dftlib/libxc.F90
```diff
@@ -433,6 +433,18 @@ subroutine libxc_input(functional_name, dft_params, tddft_params, functional)
               alpha=dft_params%cam_alpha, &
               beta=dft_params%cam_beta, &
               omega=dft_params%cam_mu)
+    case ("WB97X-D")
+      dft_params%cam_flag = .true.
+      call functional%add_functional(XC_HYB_GGA_XC_WB97X_D, 1.00_fp, &
+              alpha=dft_params%cam_alpha, &
+              beta=dft_params%cam_beta, &
+              omega=dft_params%cam_mu)
+    case ("WB97X-D3")
```
toxtran, line 442: There is no D3 part...

Alireza-Lashkaripour: Oh, so I need to add something like: dft_params%d3_flag = .true. 
I couldn't find it, do we have it?

foxtran: No, we do not have it. That is why you do not see here any functional with dispersion.

Alireza-Lashkaripour: Thank you, I will remove this part.

PR #68, opened 2025-02-28
Update libxc.F90

Alireza-Lashkaripour:


source/dftlib/libxc.F90
```diff
@@ -439,12 +439,6 @@ subroutine libxc_input(functional_name, dft_params, tddft_params, functional)
               alpha=dft_params%cam_alpha, &
               beta=dft_params%cam_beta, &
               omega=dft_params%cam_mu)
-    case ("WB97X-D3")
```
foxtran, line 442: WB97X-D should be also removed

PR #75, opened 2025-06-19
Update libxc

mohsenkor:
Update libxc to version 7.


Alireza-Lashkaripour (approved):
Apparently, in the new version of libxc, we needed to move 0.81_fp (I guess correlation mixing coefficient in B3LYP), to the beginning of the line so the rest of the values can be used correctly.

PR #78, opened 2025-07-07
fix: ixcore runtype problem

mohsenkor:
Resolved the ixcore issue during the prop run type.


Alireza-Lashkaripour (commented):
I have tested this branch and added an example for prop in the /others. It works with no problem when we run jobs with openqp directly, but when I use an external SW like [ExtOptORCA-OpenQP](https://github.com/CrespiLab/ExtOptORCA-OpenQP) the openqp log file still shows **ixcore = 1**, however when the same input used with openqp directly, it works without any problems. So I think we need more tests specially with external SW.

Alireza-Lashkaripour:
> I have tested this branch and added an example for prop in the /others. It works with no problem when we run jobs with openqp directly, but when I use an external SW like [ExtOptORCA-OpenQP](https://github.com/CrespiLab/ExtOptORCA-OpenQP) the openqp log file still shows **ixcore = 1**, however when the same input used with openqp directly, it works without any problems. So I think we need more tests specially with external SW.

Solved!

Alireza-Lashkaripour (approved):
All the test are passed! and no problem with ext. SW.

Issue #81, opened 2025-07-14
Excited state (transition) density matrices from MRSF TDDFT


amandaarcidiacono:
Dear OpenQP developers,

I have a question regarding excited state properties from MRSF TDDFT calculations, which is very interesting and useful for multi-reference systems with many excited states. 

Specifically, I am interested in extracting the density matrices of all the computed states together with all the transition density matrices between all the states, not only with the ground state. In other words, if I compute 4 singlet states, S0 (the gs) S1 S2 S3, I would like to have the **4 state density matrices** and the following **transition density matrices**:

- S0 $\rightarrow$ S1
- S0 $\rightarrow$ S2
- S0 $\rightarrow$ S3
- S1 $\rightarrow$ S2
- S1 $\rightarrow$ S3
- S2 $\rightarrow$ S3

Is this currently possible?

Thank you very much for your work and your time

Issue #80, opened 2025-07-12
MRSF-CIS wave function expression


1234zou:
Dear OpenQP developers,

Thanks for developing this excellent open source program. The MRSF-TDDFT method is very powerful for molecules with multiconfigurational/multireference characters (especially for diradicals).

I have one question which may be naive: the singlet/triplet MRSF-CIS wave function can be written as follows (no matter the MRSF code in OpenQP or in GAMESS)

![Image](https://github.com/user-attachments/assets/6099fbbe-e623-46b8-be06-fdf4e3edd8f9)

And the CI coefficients C^{i}_{a} is printed in the output file. Would you please tell me whether my understanding is correct? Thank you.

Best wishes,
Jingxiang

karmachoi:
This is not an issue of code. Please use "Discussions Section of Web."

PR #82, opened 2025-07-15
Verbose added for [tdhf]

Alireza-Lashkaripour:
A verbose keyword (verbose = 3) added for [tdhf] to print out density matrix (both state and transitions). This was requested in issues.


Alireza-Lashkaripour:
https://github.com/Open-Quantum-Platform/openqp/issues/81

1234zou:
Hi, @Alireza-Lashkaripour , may I ask you that the density matrix of each state you mentioned here, is it unrelaxed density of MRSF, or relaxed density of MRSF? Thank you.

Alireza-Lashkaripour:
Hi @1234zou,
This implementation is not correct so it doesn't print correct density matrix. We are still working on it.

1234zou:
Thank you, Alireza.

By the way, I've already deduced the unrelaxed density matrix of MRSF and use it to generate natural orbitals. I would like to contribute this code into OpenQP and GAMESS, with your (and other developers') permission. The only question is that I consider the MRSF-CIS wave function here https://github.com/Open-Quantum-Platform/openqp/issues/80, and I'm not sure whether you take the same assumption or formulae. So, would you please take a look at the formulae at that issue? If the formulae is correct, I'm sure I already deduce the correct unrelaxed density matrix of MRSF-CIS (which can also be applied to MRSF-TDDFT).

Alireza-Lashkaripour:
> Thank you, Alireza.
> 
> By the way, I've already deduced the unrelaxed density matrix of MRSF and use it to generate natural orbitals. I would like to contribute this code into OpenQP and GAMESS, with your (and other developers') permission. The only question is that I consider the MRSF-CIS wave function here #80, and I'm not sure whether you take the same assumption or formulae. So, would you please take a look at the formulae at that issue? If the formulae is correct, I'm sure I already deduce the correct unrelaxed density matrix of MRSF-CIS (which can also be applied to MRSF-TDDFT).

Dear @1234zou 
Sorry for late response. 
Thank you for your contribution, I suggest you discuss this matter in our Slack so other developers can contribute.

1234zou:
> Dear @1234zou Sorry for late response. Thank you for your contribution, I suggest you discuss this matter in our Slack so other developers can contribute.

OK, thanks for your help.

amandaarcidiacono:
Hello @Alireza-Lashkaripour, thank you for your work! I notice that this pull request is still pending, I was wondering if the new keyword is working and in that case if the update is going to be available soon. Thanks again!

h-martina:
This would be a really nice feature. Are there any news regarding the TDM?

PR #91, opened 2025-08-29
Dev

Alireza-Lashkaripour:
Add ROHF_fix for orbital reordering and remove unnecessary code in SOSCF.

This fixes the ECP issue when runtype=optimize, caused by incorrect electron counting during ECP + optimize.


karmachoi (approved):
The code is good to go!

PR #105, opened 2025-10-14
add guess=pyscf

lijingbai2009:
guess=pyscf will call pyscf to compute MOs, then use mokit to convert the MOs to the openqp format for subsequent calculations.

pyscf installation

    pip install pyscf

mokit installation

    gitclone https://gitlab.com/jxzou/mokit.git
    cd ./mokit/src
    make all -f Makefile.gnu_mkl # assume you use gfortran and mkl
    cd ./mokit/mokit
    pip install .

add the following environment variables

    export MOKIT_ROOT=$path_to_mokit/mokit
    export PATH=$MOKIT_ROOT/bin:$PATH
    export PYTHONPATH=$MOKIT_ROOT:$PYTHONPATH

due to the conflict between Numpy<2.0 and Python>3.9, you might encounter an error

    No module named 'distutils.msvccompiler'

you can modify the Numpy codes following [this](https://github.com/numpy/numpy/issues/27405)
add the following code in numpy/distutils/command/config.py

    try:
        from distutils.msvccompiler import get_build_version as get_build_msvc_version
    except ImportError:
        def get_build_msvc_version():
            return None

### DIFF
--- a/pyoqp/oqp/library/external.py
+++ b/pyoqp/oqp/library/external.py
@@ -1,0 +1,116 @@
+"""OQP external quantum chemical program"""
+import re
+import os
+import sys
+import oqp
+from oqp.utils.constants import ANGSTROM_TO_BOHR
+from oqp.periodic_table import SYMBOL_MAP, ELEMENTS_NAME
+
+try:
+    from pyscf import gto, dft, lib
+
+except ModuleNotFoundError:
+    gto = None
+    dft = None
+    lib = None
+
+try:
+    from mokit.lib import py2openqp
+
+except ModuleNotFoundError:
+    py2openqp = None
+
+pyscf_functional = {
+    "pbe0": "pbe0",
+    "b3lyp": "b3lyp",
+    "bhhlyp": "bhandhlyp",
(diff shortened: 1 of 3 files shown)


1234zou:
Great! This is an excellent update. I simply want to add some supplementary information

(i) The following two steps are not needed when one wants to compile MOKIT source code
```
cd ./mokit/mokit
pip install .
```

(ii) It is not recommended to modify Numpy code. One can simply install MOKIT via `conda`, e.g.
```
conda create -n mokit-py39 python=3.9
conda activate mokit-py39
conda install mokit -c mokit
pip install pyscf
conda deactivate
```
MOKIT and PySCF are both installed in the newly created environment `mokit-py39`. If one wants to use MOKIT and PySCF, he/she can simply run
```
source activate mokit-py39
python O2_triplet.py >O2_triplet.out 2>&1
conda deactivate
```
In such case one can install MOKIT as well as PySCF in one virtual environment, and OpenQP in another virtual environment.

mohsenkor (approved):
Nice job adding this feature.

PR #112, opened 2025-10-29
Adding an option to select the specific Python version.

karmachoi:
A new CMake option **-DOQP_PYTHON_VERSION=<version>** has been introduced, allowing users to specify the desired Python version for the build process.
With this, I think we can allow pip install dftd4 as default. 

Usage:
To build with Python 3.11, include the following flag in your CMake command:

-DOQP_PYTHON_VERSION=3.11

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,7 +171,28 @@ if(USE_LIBINT)
 endif()
 
 if(ENABLE_PYTHON)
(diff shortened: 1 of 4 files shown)


mohsenkor:
CMake already detects the Python version. Since DFTB4 is installed via pip install dftb4, I don’t see why we need to add a new CMake option to force a specific Python version.

PR #113, opened 2025-10-31
feat: default 2nd SCF converger to TRAH

mohsenkor:
This PR makes TRAH the default secondary SCF convergence method to improve stability and convergence behavior in difficult cases.


Alireza-Lashkaripour (commented):
I have tested couple of different ways of using forced_attempt and alternative_scf in this PR.
Some potential problem:
      i. When the first attempt fails, the second one starts from a different guess (not the last failed step) and it changes with each attempt. 
     ii. In some cases, when trah activates for second attempt, even though it reports that trah is failed but openqp gives "scf converged".

Alireza-Lashkaripour (approved):

PR #120, opened 2026-01-19
fix: disable SD in 2nd SCF attempt

mohsenkor:
This fixes the initial MO in the second SCF attempt, which uses the TRAH method.


karmachoi (approved):
Good work!
It is working.

Alireza-Lashkaripour (approved):

PR #124, opened 2026-04-23
feat: add excited-state electric moments

mohsenkor:
Add electric_moments_excited and electric_moments_excited_C for computing dipole, quadrupole, and octopole moments of an MRSF-TDDFT excited state using the Z-vector relaxed difference density (OQP_TD_P).

Total excited-state density is assembled as P^(0) + dP_relaxed and contracted against the existing multipole integrals. Target state is read from infos%tddft%target_state.

### DIFF
--- a/include/oqp.h
+++ b/include/oqp.h
@@ -218,6 +218,7 @@ void tdhf_mrsf_z_vector(struct oqp_handle_t *inf);
 void tdhf_mrsf_gradient(struct oqp_handle_t *inf);
 
 void electric_moments(struct oqp_handle_t *inf);
+void electric_moments_excited(struct oqp_handle_t *inf);
 void get_structures_ao_overlap(struct oqp_handle_t *inf);
 void get_states_overlap(struct oqp_handle_t *inf);
 void resp_charges(struct oqp_handle_t *inf);
--- a/include/oqp.h
+++ b/include/oqp.h
@@ -223,5 +223,6 @@ void get_structures_ao_overlap(struct oqp_handle_t *inf);
 void get_states_overlap(struct oqp_handle_t *inf);
 void resp_charges(struct oqp_handle_t *inf);
 void mulliken(struct oqp_handle_t *inf);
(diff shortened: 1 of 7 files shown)


karmachoi:
This is great. Please add Mulliken charge.

Alireza-Lashkaripour (approved):

karmachoi:
Can we have a common dipole moment routine? Right now, we have two routines for ground and excited states. They are nearly identical except density. Also in the case of MRSF-TDDFT, they are not entirely excited moments, since the lowest singlet state from MRSF-TDDFT is ground state. It is always confusing. So we better printout what state it is.

mohsenkor:
> Can we have a common dipole moment routine? Right now, we have two routines for ground and excited states. They are nearly identical except density. Also in the case of MRSF-TDDFT, they are not entirely excited moments, since the lowest singlet state from MRSF-TDDFT is ground state. It is always confusing. So we better printout what state it is.

I think it is better to have two separate routines, otherwise, we would need to introduce a new keyword.
These issues are also fixed in this pull request:
1. RESP charges likely have an uninitialized-size allocation bug. In resp.F90 (line 111), q0(nat) and alpha(nat) are allocated before nat is assigned at line 138. If infos%control%esp == 2, this can allocate with undefined size.

2. MRSF debug printing is always enabled. In tdhf_mrsf_energy.F90 (line 317), if(debug_mode.or..true.)then makes debug output unconditional. That will hurt large batch runs and MPI/OpenMP logs.

3. UMRSF spin-pair coupling has a likely missing block. In tdhf_mrsf_energy.F90 (line 490), only line 491 is guarded by the if; lines 492-498 always scale components 2-8 by spc_coov / hfscale. If that was intended only when spc_coov /= hfscale, it needs then … endif.


5. Python barray() parses "False" as True. In oqpdata.py (line 25), bool("False") is true in Python. If boolean arrays are added for property options later, this will silently misread inputs.

PR #128, opened 2026-05-18
SOC (1e and 2e cases)

VladimirMakhnev:
Added: SOC calculations.
 
The calculation procedure:
SCF -- MRSF (singlet) -- MRSF (triplet) -- SOC
(see runfunc.py)

Keywords: 
- new runtype: "runtype = soc"
- new input keyword: "soc_2e". Two modes supported:
   "soc_2e = 0" -- only 1e SOC calculation
   "soc_2e = 1" -- 1e and 2e SOC calculation
   
New input folder -- "SOC" with examples of thymine and CH3Br. 

New module: "soc_mrsf.F90".

PR #166, opened 2026-06-01
SOC MRSF (1e,2e)

VladimirMakhnev:
The modified SOC realisation (see #128 ) according the comments.


karmachoi:
Pushed a small correctness fix to this branch (commit b0e76602): **2e SOC density screening was dropping exchange contributions.**

### Bug
`soc2e_rys_compute` screened quartets with `dabmax = max|D(K,L)|` — the ket-block density only. That bound is valid for the Coulomb term, but the exchange terms in `compute_soc2e_ao` contract the cross blocks D(I,L), D(J,L), D(I,K), D(J,K). Whenever D(K,L) ≈ 0 while a cross block is finite, the whole quartet was skipped and its exchange contribution silently lost.

For a spherical atom the s–p density blocks are *exactly* zero, so **all** exchange contributions vanished — element-by-element the 2e SOC matrix equalled exactly −2·vj. General molecules have dense blocks, so the loss there is only ~1e-6, which is why molecular validation never caught it; high-symmetry atoms are the pathological case.

### Fix
`dabmax` now bounds max|D| over all five density blocks used by the contraction (KL, IL, JL, IK, JK). No change to the integrals or the contraction algebra; the dead `den_kl` scratch array is removed.

### Validation
| BHHLYP/cc-pVTZ + DKH2 | before | after | reference |
|---|---|---|---|
| C 3P SOC spacing (cm-1) | 22.19 | **16.28** | 16.4 exp / published |
| C screening ratio ζ(1e+2e)/ζ(1e) | 0.734 | **0.538** | 0.535 (independent PySCF Eq. 8) |
| Si screening ratio | 0.854 | **0.772** | 0.773 (PySCF Eq. 8) |

The bundled H2O / CH3Br reference JSONs still pass (soc_eval shifts ~2e-6), so no reference regeneration is needed. Cross-checked against an independent SOMF implementation (PySCF `int2e_p1vxp1`, vj − 3/2 vk − 3/2 vkᵀ, same density, verified AO permutation); a line-by-line Python emulation of the `compute_soc2e_ao` accumulation reproduces the Eq.-8 result to machine precision — i.e. the contraction and the closed-shell density convention were already correct, only the screening was wrong. Full evidence matrix in `docs/soc2e_screening_analysis.md`.

### One open question for the authors
After the fix, OpenQP agrees with the mean-field Eq.-8 result to <0.5% for both C and Si. But Si then gives 69.2 cm⁻¹ vs the published 75.7 / experimental 77.1. Since the mean-field physics now cross-checks exactly, the residual gap suggests the published GAMESS numbers may have used the *full* Breit–Pauli 2e operator (HSO2/HSO2FF) rather than the mean-field screening. Could you confirm which 2e treatment the published values used? cc @Konstantin

PR #178, opened 2026-06-06
fix: dangling ECP CFFI buffers corrupt nuclear repulsion for large systems

karmachoi:
## Summary

Fixes a use-after-free in the Python→Fortran ECP interface that has silently corrupted the nuclear repulsion energy for large systems since the ECP commit (d1398dfe, Feb 2025).

`set_ecp_data()` built the `ecp_zn`/`ecp_am` C pointers from **temporary** numpy arrays:

```python
self.mol.data["ecp_zn"] = ffi.cast("int*", ffi.from_buffer(np.array(self.ecp["ecp_electron"], dtype=np.int32)))
```

`ffi.from_buffer` does not hold a reference to its argument, so the temporary is freed as soon as the statement ends and `mol.data` is left with a dangling pointer. `oqp_append_ecp` (basis_api.F90) is called unconditionally — even with no ECP — and copies `natom` integers from that pointer *before* the `element_id == 0` early return. The garbage propagates to `basis%ecp_zn_num`, and the SCF driver computes the nuclear repulsion from `zn - ecp_zn_num`, turning the total energy into a huge garbage constant while the Fock/density/DIIS path stays correct.

Small systems survived by luck (the tiny freed block was rarely reused). Large systems failed deterministically — e.g. the 1BNA benchmark (758 atoms, RHF/STO-3G, 2790 AOs, charge −22):

```
 Iter         Energy            Delta E         Int Skip     DIIS Error
    1  ***************** ***************** 1237063183430 **************
    2  ***************** ***************** 1237024131088     0.41034169
```

The neighbouring arrays (`n_expo_array`, `expo_array`, `coef_array`, `r_expo_array`, `coord_array`) were already bound to named locals; `ecp_am`/`ecp_zn` now follow the same pattern.

## Verification

- 1BNA RHF/STO-3G iteration 1 restored to `-31414.4194501916` (DIIS `0.71894225`), matching the pre-ECP reference log (`-31414.4194501924` / `0.71894225`) to 12 significant figures.
- H2O RHF/STO-3G smoke unchanged: `-74.9631468000`.
- New regression test monkeypatches `oqp.append_ecp`, churns the heap, and reads the buffer back: on the unfixed code it reads recycled heap (the churn fill value) instead of the ECP electron counts; with the fix it reads the correct values. A source-level check also bans the `ffi.from_buffer(np.array(` pattern in set_basis.py.

```
tests/test_set_basis_ecp_buffer_lifetime.py  2 passed
```

### DIFF
--- a/pyoqp/oqp/library/set_basis.py
+++ b/pyoqp/oqp/library/set_basis.py
@@ -194,8 +194,14 @@ class BasisData:
         self.mol.data["ecp_rex"] = ffi.cast("int*", ffi.from_buffer(r_expo_array))
 
         coord_array = np.array(self.ecp["coord"], dtype=np.float64)
-        self.mol.data["ecp_am"] = ffi.cast("int*", ffi.from_buffer(np.array(self.ecp["ang"], dtype=np.float64)))
-        self.mol.data["ecp_zn"] = ffi.cast("int*", ffi.from_buffer(np.array(self.ecp["ecp_electron"], dtype=np.int32)))
+        # Keep the numpy arrays alive in named variables until append_ecp has
+        # consumed them: ffi.from_buffer does not hold a reference, so passing
+        # a temporary leaves mol.data with a dangling pointer (garbage
+        # ecp_zn_num -> garbage nuclear repulsion for large systems).
+        ecp_am_array = np.array(self.ecp["ang"], dtype=np.float64)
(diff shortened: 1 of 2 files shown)


karmachoi:
Re-opening from fork per project workflow.

PR #179, opened 2026-06-06
fix: large-system SCF failure (ECP dangling buffers + C-DIIS selection + log format)

karmachoi:
## Summary

Fixes the large-system SCF failure where RHF/STO-3G on the 1BNA benchmark (758 atoms, 2790 AOs, charge −22) printed `*****` energies from iteration 1 and never converged, despite working in earlier releases. Three independent regressions were involved; each has a fix and a test.

### 1. Dangling ECP CFFI buffers (root cause) — `pyoqp/oqp/library/set_basis.py`

`set_ecp_data()` built the `ecp_zn`/`ecp_am` C pointers from **temporary** numpy arrays:

```python
self.mol.data["ecp_zn"] = ffi.cast("int*", ffi.from_buffer(np.array(self.ecp["ecp_electron"], dtype=np.int32)))
```

`ffi.from_buffer` does not hold a reference, so the temporary is freed immediately and `mol.data` keeps a dangling pointer. `oqp_append_ecp` (basis_api.F90) is called unconditionally — even with no ECP — and copies `natom` integers from that pointer *before* the `element_id == 0` early return, poisoning `basis%ecp_zn_num`. The SCF driver computes nuclear repulsion from `zn − ecp_zn_num`, so heap reuse turned the total energy into a huge garbage constant while Fock/density/DIIS stayed correct. Small systems survived by luck (tiny freed block rarely reused); large systems failed deterministically. Introduced by the ECP commit d1398dfe (Feb 2025). Fix: bind the arrays to named locals so they outlive `append_ecp`.

### 2. SCF converger selection drops C-DIIS — `source/scf_converger.F90`

`scf_conv_select` reused its loop counter `i` in a dead SOSCF-detection scan added by c0ccabae (Jun 2025); the computed `use_soscf_now` flag was never read. After the scan, `i` always equals `nconv+1`, so the "promote from SD to the real converger after the first step" rule (`if (i == 0 .and. self%step > 0)`) could never fire. Any iteration with DIIS error above the first threshold (default 2.0) silently fell back to plain SD instead of C-DIIS, and SCF bounced/diverged. Fix: delete the dead scan, restoring pre-c0ccabae selection.

### 3. Iteration table hides oversized values — `source/scf.F90`

The table used bare `f17.10`/`f14.8` descriptors, so any value too large for its column printed as `*****` — which is exactly what hid regression #1's magnitude. The energy/ΔE and error/gradient columns now route through helpers that keep the usual fixed-point form (and column width) while the value fits and fall back to same-width scientific notation otherwise. Normal output is byte-for-byte unchanged; NaN prints as `NaN`.

## Verification

Rebuilt from this branch (gfortran-15, Release, OpenMP) and re-run:

- **H2O RHF/STO-3G**: `-74.9631468000` (unchanged).
- **1BNA RHF/STO-3G**: energies finite from iteration 1, and **C-DIIS now engages at iteration 2** (regression #2 fixed) instead of falling back to SD. With the unfixed converger the trajectory diverged to *positive* ~+20000 Ha by iteration 5.
- New tests: `tests/test_set_basis_ecp_buffer_lifetime.py` (heap-churn use-after-free check + source-pattern ban), `tests/test_scf_iteration_log_format.py` (formatter routing + same-width fallback). All pass.

```
tests/test_set_basis_ecp_buffer_lifetime.py  2 passed
tests/test_scf_iteration_log_format.py       2 passed
```

(Supersedes #178, re-opened from fork per project workflow.)

### DIFF
--- a/pyoqp/oqp/library/set_basis.py
+++ b/pyoqp/oqp/library/set_basis.py
@@ -194,8 +194,14 @@ class BasisData:
         self.mol.data["ecp_rex"] = ffi.cast("int*", ffi.from_buffer(r_expo_array))
 
         coord_array = np.array(self.ecp["coord"], dtype=np.float64)
-        self.mol.data["ecp_am"] = ffi.cast("int*", ffi.from_buffer(np.array(self.ecp["ang"], dtype=np.float64)))
-        self.mol.data["ecp_zn"] = ffi.cast("int*", ffi.from_buffer(np.array(self.ecp["ecp_electron"], dtype=np.int32)))
+        # Keep the numpy arrays alive in named variables until append_ecp has
+        # consumed them: ffi.from_buffer does not hold a reference, so passing
+        # a temporary leaves mol.data with a dangling pointer (garbage
+        # ecp_zn_num -> garbage nuclear repulsion for large systems).
+        ecp_am_array = np.array(self.ecp["ang"], dtype=np.float64)
+        ecp_zn_array = np.array(self.ecp["ecp_electron"], dtype=np.int32)
+        self.mol.data["ecp_am"] = ffi.cast("int*", ffi.from_buffer(ecp_am_array))
+        self.mol.data["ecp_zn"] = ffi.cast("int*", ffi.from_buffer(ecp_zn_array))
         self.mol.data["ecp_coord"] = ffi.cast("double*", ffi.from_buffer(coord_array))
 
         oqp.append_ecp(self.mol)
--- a/pyoqp/oqp/molecule/oqpdata.py
+++ b/pyoqp/oqp/molecule/oqpdata.py
@@ -57,7 +57,7 @@ OQP_CONFIG_SCHEMA = {
         'd4': {'type': bool, 'default': 'False'},
     },
     'guess': {
-        'type': {'type': string, 'default': 'sap'},
+        'type': {'type': string, 'default': 'huckel'},
         'file': {'type': str, 'default': ''},
         'file2': {'type': str, 'default': ''},
         'save_mol': {'type': bool, 'default': 'False'},
--- a/source/scf.F90
+++ b/source/scf.F90
@@ -706,19 +706,22 @@ contains
       !----------------------------------------------------------------------------
       ! Print iteration information
       if (infos%control%pfon) then
-        write(IW,fmt="(4x,i4.1,2x,f17.10,1x,f17.10,1x,i13,1x,f14.8,5x,f5.3,5x,a,5x,a,f9.2)") &
-              iter, energy%etot, energy%etot - e_old, nschwz, diis_error, vshift, &
+        write(IW,fmt="(4x,i4.1,2x,a17,1x,a17,1x,i13,1x,a14,5x,f5.3,5x,a,5x,a,f9.2)") &
+              iter, fmt_real17(energy%etot), fmt_real17(energy%etot - e_old), nschwz, &
+              fmt_real14(diis_error), vshift, &
               trim(conv_res%active_converger_name), "Temp:", pfon%temp
         write(IW,fmt="(100x,a,f9.2)") "Beta:", pfon%beta
       elseif (infos%control%converger_type == scf_bfgs) then
-        write(IW,'(4x,i4.1,2x,f17.10,1x,f17.10,1x,i13,1x,f14.8,1x,f14.8,5x,f5.3,5x,a)') &
-              iter, energy%etot, energy%etot - e_old, nschwz, rms_grad, rms_dp, vshift, &
+        write(IW,'(4x,i4.1,2x,a17,1x,a17,1x,i13,1x,a14,1x,a14,5x,f5.3,5x,a)') &
+              iter, fmt_real17(energy%etot), fmt_real17(energy%etot - e_old), nschwz, &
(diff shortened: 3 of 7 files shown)


karmachoi:
Added a second commit: the SCF iteration table now falls back to same-width scientific notation when a value cannot fit its column (previously printed `*****`, which hid the garbage magnitude during this bug's diagnosis). Normal output is byte-for-byte unchanged; NaN prints as `NaN`. Verified with a standalone gfortran check of the formatter widths.

karmachoi:
Pushed a third commit (cfb27b5f) and expanded the PR to cover all three regressions found while tracing the 1BNA failure. Verified against a fresh gfortran-15 build: H2O unchanged (-74.9631468000), 1BNA now finite from iter 1 with C-DIIS correctly engaging at iteration 2 (was diverging to +20000 Ha before). Full convergence run to -31518.4587 in progress.

karmachoi:
Added a 4th commit (bbd581ba): restore Hückel as the default initial guess. Verified on the rebuilt runtime — with Hückel, 1BNA reproduces the pre-2025 reference trajectory iteration-for-iteration (iter 1 -31414.41945 / DIIS 0.71894225, same SD→C-DIIS method sequence), converging to -31518.45873. SAP also converges now that the buffer/converger bugs are fixed but oscillates early and needs a DIIS reset, so Hückel is the safer default for large anions.

karmachoi (commented):
## Hermes review

Reviewed PR #179 at head `bbd581ba16b31d7780b39ee4fa565dab518fa6b4`.

Overall this looks reasonable. The ECP buffer-lifetime fix matches the reported failure mode: `append_ecp` copies `ecp_zn` before returning on `element_id == 0`, so keeping the NumPy arrays alive through the call is necessary. The C-DIIS selector cleanup appears to restore the intended "SD only on first step, then promote" behavior. The SCF formatting change is localized and keeps column widths while avoiding `*****` output for oversized values.

Validation run locally:
- `git diff --check origin/main...HEAD` passed.
- `python3.11 -m py_compile` on changed Python files passed.
- `python3.11 -m pytest -q tests/test_default_guess_huckel.py tests/test_scf_iteration_log_format.py tests/test_set_basis_ecp_buffer_lifetime.py` passed: `5 passed in 0.29s`.

Notes / suggestions, non-blocking:
1. `tests/test_set_basis_ecp_buffer_lifetime.py` verifies `ecp_zn` functionally, but the PR also fixes `ecp_am` lifetime. Consider adding a small ECP-active fixture that captures `mol.data["ecp_am"]` as well, so the second fixed buffer has value-level coverage rather than only the source-pattern assertion.
2. In source-tree/uninstalled mode, `PYTHONPATH=$PWD/pyoqp python3.11 -m pytest ...` hit `FileNotFoundError` importing `oqp` because `pyoqp/oqp/include/oqp.h` is absent before build/install. CI appears to install/build first, so this is not blocking, but if source-tree pytest is intended, consider skipping this test explicitly when generated headers/native libraries are unavailable.
3. `test_scf_iteration_log_format.py` is source-pattern based. That is useful, but a lightweight behavioral check of the formatter output for a boundary/oversized value would make the regression test stronger if there is a practical Fortran-test hook.

Verdict: comment-only; no blocking issues found from the inspected diff and focused checks.

karmachoi:
Addressed the review suggestions in commit `0c3c29a`:

- Added functional `ecp_am` buffer-lifetime coverage for an ECP-active handoff path, in addition to the existing no-ECP `ecp_zn` check.
- Made the `oqp` import in the buffer-lifetime test skip cleanly when a source-tree checkout lacks generated native build artifacts such as `include/oqp.h`.
- Added a lightweight behavioral formatter test that compiles a tiny Fortran program and verifies fixed-width/overflow values print as numbers, not `*****`, while preserving the intended column widths.

Focused validation:
- `python3.11 -m py_compile tests/test_set_basis_ecp_buffer_lifetime.py tests/test_scf_iteration_log_format.py`
- `python3.11 -m pytest -q tests/test_default_guess_huckel.py tests/test_scf_iteration_log_format.py tests/test_set_basis_ecp_buffer_lifetime.py` → `7 passed in 0.90s`
- `PYTHONPATH=$PWD/pyoqp python3.11 -m pytest -q tests/test_set_basis_ecp_buffer_lifetime.py` → `1 passed, 2 skipped in 0.19s`

karmachoi (commented):
## Codex review

No blocking issues found.

Codex reviewed only `origin/main...HEAD` at `0c3c29a`, ignoring untracked local files. The ECP buffer lifetime fix, Hückel default restore, SCF formatter fallback, and converger-selection cleanup look consistent with the changed Python/Fortran integration points.

Note: Codex did not run tests because it ran in a read-only sandbox. Focused tests were run separately after the fix:
- `python3.11 -m py_compile tests/test_set_basis_ecp_buffer_lifetime.py tests/test_scf_iteration_log_format.py`
- `python3.11 -m pytest -q tests/test_default_guess_huckel.py tests/test_scf_iteration_log_format.py tests/test_set_basis_ecp_buffer_lifetime.py` → `7 passed in 0.90s`
- `PYTHONPATH=$PWD/pyoqp python3.11 -m pytest -q tests/test_set_basis_ecp_buffer_lifetime.py` → `1 passed, 2 skipped in 0.19s`

PR #180, opened 2026-06-06
Huckel guess: refactor + up to ~100x speedup (SVD projection, DSYEVD, S^(-1/2) cache)

karmachoi:
## Summary

Refactoring and optimization of the extended Huckel initial guess (`huckel` / `modhuckel`), plus shared-infrastructure improvements that benefit all guesses and the SCF setup.

**Headline numbers** (Huckel guess wall time, C32H66 / cc-pVTZ, nbf = 1884, Apple M-series, ILP64 OpenBLAS):

| Configuration | Time | Speedup |
|---|---|---|
| `main` | 10.7 s | — |
| this PR, first guess at a geometry (cold) | 0.35 s | **~30x** |
| this PR, S^(-1/2) cache warm | 0.11 s | **~100x** |

At nbf = 956 (C16H34 / cc-pVTZ): 1.52 s → 0.088 s (~17x). The speedup grows with system size: the dominant cost dropped from four O(N³) eigendecompositions to a single cached one plus O(N²·n_occ) work. All changes are numerically equivalent — MO norms match `main` to all printed digits in every benchmark, and the full example suite (175 tests: SCF, HF, DFT, other, MRSF-TDDFT) passes against unchanged reference JSONs at every commit.

## What changed

1. **Driver deduplication** — `guess_modhuckel.F90` was a 157-line near-copy of `guess_huckel.F90`; both C entry points now share one `guess_huckel_driver(infos, modified)`. C ABI and Python interface unchanged.
2. **Dead code removal & latent bug fixes** — unused `s2`/`wrk2`/`scr` allocations; uninitialized `core(:)` read for dummy atoms; uninitialized `err` on non-root MPI ranks; LAPACK workspace query with unallocated `tau`; undersized `lwork` for `dormqr(side='R')`; `orb_to_dens` occupation indexing (`x(m)` → `x(i)`); clean abort when the MINI basis file cannot be read.
3. **Thin SVD in the corresponding-orbital projection** — only the `nmo` leading eigenvectors of `D^T D` are ever consumed (`schmd` regenerates the rest), so a thin `dgesvd` of the small `(nmo × nrest)` matrix replaces the O(nrest³) diagonalization.
4. **Householder rotation** — the projection rotation `va·Q` is applied directly from the QR reflectors (`dgeqrf`+`dormqr`) at O(nbf·nrest·nmo) instead of forming Q explicitly (`dorgqr`) and multiplying at O(nbf·nrest²).
5. **Faster LAPACK drivers** — `diag_symm_full`: DSYEV → DSYEVD; `matrix_invsqrt`: packed DSPEV → unpack + DSYEVD (packed eigensolvers cannot use blocked level-3 BLAS and are an order of magnitude slower at large nbf). Benefits SCF and TDHF as well.
6. **S^(-1/2) cache** (`OQP::QMAT`) — every guess and the SCF setup computed the same canonical orthogonalizer independently. It is now computed once and reused; `int1e` invalidates the record whenever the overlap matrix is recomputed (new geometry/basis), so staleness is impossible.
7. **Huckel orbital energies exported to `E_MO_A/B`** — previously zeros until the first SCF diagonalization; makes `guess.swapmo` and orbital diagnostics meaningful before SCF.
8. **BITSY damping study (documented in-code)** — on a 10-molecule HF/cc-pVDZ set, flat BITSY ∈ {0, 0.01, 0.05, 0.1, 0.2} changes SCF iteration counts by ≤1 everywhere except SiCl4; the energy-dependent damping suggested by the old TODO comment is severely counterproductive (SiCl4: 29 → 73 iterations). The long-standing default 0.05 is kept and the TODO replaced by the measured results.

## macOS BLAS guidance (important)

OpenQP requires ILP64 BLAS/LAPACK. On macOS, **no system or Homebrew package provides ILP64 with standard Fortran symbol names**, so CMake silently falls back to the bundled netlib reference BLAS — single-threaded and statically embedded into `liboqp` — which costs roughly 3–30x in guess/linear-algebra performance. Recommendations:

- **Use OpenBLAS built from source with `INTERFACE64=1`** (and a `libopenblas64.dylib` symlink so CMake's ILP64 search finds it). This is the fastest verified option.
- **Apple's native Accelerate framework is measurably slower**: its ILP64 LAPACK (`$NEWLAPACK$ILP64` symbols, bridged to gfortran via assembly trampolines for testing) gave 0.55 s vs OpenBLAS's 0.35 s on the nbf = 1884 benchmark, because Accelerate does not multithread the reduction steps the way OpenBLAS does. Its ILP64 BLAS is CBLAS-only and cannot be linked from Fortran directly at all.
- Consider emitting a CMake **warning** when the netlib fallback is selected, so users know they are getting reference-BLAS performance.

## Validation

- 175 example-suite tests (SCF/HF/DFT/other/MRSF-TDDFT) match reference JSONs at every commit in this PR.
- MO norms identical to `main` to all printed digits in all benchmarks (the optimizations are exact reformulations, not approximations).


karmachoi (commented):
## Hermes PR Review

Verdict: no actionable findings from inspect-only review.

Scope reviewed:
- Huckel/modHuckel driver deduplication and MINI-basis projection path.
- Thin SVD + Householder QR corresponding-orbital projection rewrite.
- `matrix_invsqrt` packed eigensolver -> full DSYEVD path.
- `OQP::QMAT` cache creation/reuse/invalidation via `int1e`.
- Guess call sites for hcore/huckel/modhuckel/minao/sap and SCF reuse.

Local checks run on PR head `e3c012f53720d7691da1dc25e702aa2f27ff75fe`:
- `git diff --check origin/main...HEAD` — passed.
- `python3.11 -m pytest -q tests/test_default_guess_huckel.py tests/test_advanced_guess.py tests/test_advanced_guess_examples.py` — passed: 6 tests + 5 subtests.
- Non-blocking Codex read-only review focused on the Fortran changes — no actionable findings.

Build note:
- I attempted a clean local CMake build with `ENABLE_PYTHON=OFF`, but it was blocked before OpenQP sources by an external `libtagarray` configure error (`target_compile_check_add_flags Function invoked with incorrect arguments`). I did not count that as a PR regression because it occurred in external dependency configuration before compiling the changed OpenQP sources.

CI was still pending when checked.

PR #187, opened 2026-06-07
Remove libint2 ERI backend; native Rys quadrature is the single ERI engine

karmachoi:
## Summary

Removes the optional libint2 two-electron integral backend. The native Rys
quadrature code becomes the single ERI engine (with the existing rotated-axis
fast path for L ≤ 2), matching how the project already ships everywhere:
the README build recipes, the Dockerfile, the pip/scikit-build configuration,
and both the GitHub and GitLab CI pipelines all build with `USE_LIBINT=OFF`.

## Motivation

- The libint backend was unexercised by CI, so it could (and did) rot silently:
  on macOS arm64 with current GCC the generated libint kernels are miscompiled
  and **every d/f-shell ERI calculation segfaults** in `int2_twoei` (cc-pVTZ
  examples crash; 6-31G works). Bisection across recent main showed this at all
  source versions whenever `USE_LIBINT=ON`.
- It doubled the dispatch logic in `int2.F90` (init/cleanup plumbing,
  `erieval` buffers, three-way backend selection in two places).
- When enabled it added an external tarball download and ~25 minutes of C++
  compilation to every clean build.

## Changes

- Drop the `USE_LIBINT` CMake option, the `libint2` ExternalProject, the
  include/link wiring, and the `FortranCInterface_VERIFY(CXX)` check.
- Delete `source/integrals/int_libint.F90` and `source/integrals/libint_f.F90`.
- `int2.F90`: remove the libint dispatch branches and buffers;
  `shellquartet`/`ints_exchange` now select rotated-axis for L ≤ 2 and Rys
  otherwise — numerically identical to the shipped `USE_LIBINT=OFF` builds.
- Scrub `-DUSE_LIBINT=OFF` flags from README, Dockerfile, pyproject.toml, and
  the CI configs (it is now meaningless), and drop the stale
  `external/libint` CI cache entries.

## Validation (macOS arm64, gfortran-15, ILP64 OpenBLAS)

- 12 example directories (79 tests) pass through the official
  `--run_tests` runner: SCF 31/31, DFT, HF, TDDFT, SF-TDDFT, TDHF, TRAH,
  XAS, ECP 9/9, UMRSF, MRSF-TDDFT, OPT 11/11.
- cc-pVTZ d/f coverage (the L > 2 range libint used to handle):
  `h2o-2_rhf_cc-pvtz_hf`, `h2o_{rhf,rohf,uhf-s}_cc-pvtz_b3lypv5` — 4/4 pass.
- `pytest tests/` matches a pristine upstream-main build bit-for-bit in
  outcome: 162 passed and the identical set of 16 pre-existing,
  environment-specific failures on both builds (verified by diffing the
  failure lists) — i.e. zero regressions from this change.

## Benchmarks (Rys, this PR vs libint, upstream main `c82947b` + `USE_LIBINT=ON`)

RHF/direct SCF, macOS arm64 (M-series), gfortran-15, 8 OpenMP threads, ILP64 OpenBLAS.
Water clusters chosen so that L > 2 quartets dominate (libint only engages for `max_am > 2`).

| system | basis (nbf) | Rys — this PR | libint — main | energy agreement |
|---|---|---|---|---|
| (H2O)8 | cc-pVDZ (192) | **36.9 s** (11 iters) | 204.7 s (11 iters) | −608.23468366 Ha, identical to all printed digits |
| (H2O)2 | cc-pVTZ (116) | **5.4 s** (10 iters) | **SIGSEGV** in `int2_twoei` | — |
| (H2O)4 | cc-pVTZ (232) | **50.4 s** (10 iters) | **SIGSEGV** in `int2_twoei` | — |

- d-shell case: Rys is ~5.5× faster *as integrated by the project's build system* — the
  libint external project does not forward an optimization level (`CMAKE_BUILD_TYPE` is
  empty in the sub-build), so libint has always shipped unoptimized via `USE_LIBINT=ON`.
  This is an out-of-the-box comparison, not a claim against a hand-tuned libint build.
- f-shell cases: the libint backend segfaults on this platform (miscompiled generated
  kernels, reproducible at every recent upstream revision), while the Rys path runs them
  routinely — the original motivation for this PR.
- Where both backends run, energies and iteration counts are identical.


karmachoi:
Closing: this was over-scoped — the intent was to keep NMR on the native Rys path only, not to remove the libint2 backend project-wide. The macOS arm64 d/f segfault and the benchmark data in the description remain valid information if backend removal is ever reconsidered.

PR #191, opened 2026-06-07
Direct-SCF integral/Fock performance: sequential-BLAS threading, low-memory atomic Fock, adaptive quartet dispatch

karmachoi:
## Summary

Three independent, low-risk performance improvements to the direct-SCF integral/Fock path, plus the threading defaults that make them effective. SCF energies are preserved (bit-for-bit, or to FP summation order), and behavior-changing pieces are off by default / opt-in.

1. **Sequential-BLAS + worker-stack threading defaults** (`pyoqp/oqp/pyoqp.py`)
2. **Density-sparsity-gated low-memory Fock accumulator** (`source/integrals/int2.F90`)
3. **Adaptive (cost-sorted) shell-quartet dispatch** (`source/integrals/int2.F90`)

Benchmarks are from **dedicated, exclusive Intel Xeon nodes (uncontended), MKL ILP64**.

---

## 1. Sequential-BLAS threading defaults

OpenQP parallelizes the 2e integral/Fock build with OpenMP. If the linked BLAS *also* spawns threads inside those OpenMP regions, the cores are oversubscribed (`OMP_threads × BLAS_threads`) and the build stalls. Before native libraries load we now `setdefault`:

```
OPENBLAS_NUM_THREADS=1  MKL_NUM_THREADS=1  BLIS_NUM_THREADS=1  VECLIB_MAXIMUM_THREADS=1
OMP_STACKSIZE=256M  GOMP_STACKSIZE=256M
```

(plus `OMP_NUM_THREADS=16` on macOS to stay on performance cores). These are `setdefault`, so an explicit user environment still wins. `OMP_STACKSIZE` also fixes a libgomp worker-stack startup crash on macOS/arm64.

**Hardwired, env-proof variant.** Because the env var above can be re-introduced accidentally (e.g. a stray `OPENBLAS_NUM_THREADS` in a shell profile), the 2e build *also* forces BLAS to a single thread at the code level, via the BLAS library's own runtime setter (`openblas_set_num_threads` / `MKL_Set_Num_Threads` / BLIS, through the existing `blas_thread` module): it saves the thread count, sets 1 for the duration of `int2_twoei`, and restores it on exit. This cannot be overridden by the environment, and it is **scoped** — diagonalisation/TDDFT/Hessian linear algebra outside the integral build keep full BLAS threading. It mirrors the pattern already used in `dft_gridint`. No-op for reference BLAS / Apple Accelerate (no runtime setter).

**Impact** — caffeine 6-31G(d,p), Rys engine, dedicated 56-core Xeon, sequential BLAS:

| threads | wall (s) | speedup | parallel eff. |
|--------:|---------:|--------:|:-------------:|
|   1     | 1452.2   |  1.0x   |   --          |
|   4     |  367.4   |  3.95x  |  99%          |
|   8     |  187.5   |  7.74x  |  97%          |
|  16     |   97.5   | 14.9x   |  93%          |
|  28     |   59.9   | **24.2x** | **86%**     |

Energy -676.34107 (== Rys) at every thread count. With BLAS oversubscription the same 28-thread run was ~1.5x slower and appeared to plateau past 16 threads; the plateau was oversubscription, not the integral code.

---

## 2. Density-sparsity-gated low-memory Fock accumulator

The Fock build keeps one full Fock copy **per thread** (`fockdim*nfocks*nthreads*8 B`) and reduces at the end -- fast, but memory grows with thread count and becomes prohibitive for large bases (5000 bf x 28 threads ~ 2.8 GB; 10000 bf ~ 11 GB).

This adds an opt-in **single shared Fock with `!$omp atomic` accumulation** (`O(fockdim)` memory, ~`Nthreads`x smaller). Per-integral atomics only contend cheaply when few quartets survive screening, so it auto-engages **only when both**:
- replicated buffers would exceed `OQP_FOCK_MEM_MB` (default 4096), **and**
- the shell density is sparse (significant-pair fraction < `OQP_FOCK_SPARSITY`, default 0.5).

Dense/compact systems keep the fast replicated path. `OQP_FOCK_ATOMIC=1/0` forces the mode; the chosen mode, MB saved, and `frac_sig` are printed. Covers RHF and UHF/ROHF; the cross-thread reduction is skipped in atomic mode.

**Validation** -- (H2O)32 / cc-pVDZ (800 bf), 28 threads, replicated vs forced-atomic:

| mode | peak RSS | E_RHF |
|------|---------:|------:|
| replicated | 433 MB | -4864.0529276082 |
| atomic     | 400 MB | -4864.0529276082 |

Energy **bit-exact**. Analytic Fock-memory saved `(nthreads-1)*fockdim*nfocks*8`: ~96% -- **2.7 GB at 5000 bf, 10.8 GB at 10000 bf** (28 threads). Forced-atomic is slower on dense systems (atomic contention), which is why the gate restricts it to large+sparse -- the regime where memory is the binding constraint and contention is low.

---

## 3. Adaptive (cost-sorted) shell-quartet dispatch

The flat shell-pair work list driving the `schedule(dynamic)` 2e loop is now ordered by **descending estimated cost** -- `Schwarz(i,j) * nbf_i * nbf_j * i` -- via an O(n) counting sort. Dynamic scheduling then hands out the expensive, density-significant quartets first and lets cheap/screened diffuse pairs fill the tail, reducing the end-of-region load-imbalance barrier wait at high thread counts. Reordering changes only work distribution, never results.

---

## Correctness & compatibility

- SCF energies unchanged: H2O 6-31G(d,p) = -76.0235728233 (== Rys) in every mode; caffeine and (H2O)32 as above.
- Default behavior preserved (atomic Fock gated/opt-in; threading vars `setdefault`; dispatch reordering result-invariant).
- No new build dependencies.

## Reproduce

Sequential-BLAS scaling is automatic. Low-memory Fock on a large sparse system: `OQP_FOCK_ATOMIC=1 OMP_NUM_THREADS=28 openqp big.inp` (or rely on the auto-gate for >4 GB replicated + sparse density).

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,34 @@ For OpenMP and MPI run:
 mpirun -np number_of_mpi openqp any_example_file.inp
 ```
 
+#### Performance and threading
+
+OpenQP parallelizes the two-electron integral / Fock build with OpenMP. A few
+defaults and knobs keep it efficient:
+
+- **Sequential BLAS inside the integral build (automatic).** A threaded BLAS
+  keeps its own worker pool that would oversubscribe the cores against the
+  OpenMP integral threads. OpenQP forces BLAS to a single thread for the
+  duration of the integral build (via the BLAS library's own runtime setter, so
+  it cannot be re-introduced by a stray `OPENBLAS_NUM_THREADS`/`MKL_NUM_THREADS`)
+  and restores full BLAS threading afterward for diagonalization and other
+  BLAS-heavy phases. The frontend also sets conservative `*_NUM_THREADS=1` and
+  `OMP_STACKSIZE` defaults (via `setdefault`, so an explicit environment wins).
+  Measured ~1.5x faster at 24-28 threads on a dedicated node; near-ideal scaling
+  to the performance-core count.
+
+- **Low-memory Fock accumulator (automatic for large, sparse systems).** The
+  Fock build normally keeps one Fock copy per thread
+  (`fockdim x nfocks x nthreads`); for large bases on many cores this can reach
+  several GB. OpenQP can instead accumulate into a single shared Fock with atomic
+  updates (~`Nthreads`x less memory). It auto-engages only when the replicated
+  buffers would exceed a cap **and** the density is sparse (so atomic contention
+  is negligible). Controls:
+  - `OQP_FOCK_MEM_MB` — replicated-memory cap in MB before switching (default `4096`).
+  - `OQP_FOCK_SPARSITY` — significant shell-pair fraction below which the density
+    is considered sparse (default `0.5`).
+  - `OQP_FOCK_ATOMIC=1`/`0` — force the shared-atomic / replicated mode regardless.
+
 ### Detailed Documentation
 
 For more in-depth information, visit:
--- a/pyoqp/oqp/pyoqp.py
+++ b/pyoqp/oqp/pyoqp.py
@@ -9,6 +9,38 @@ import sys
 import time
 import argparse
 from signal import signal, SIGINT, SIG_DFL
+
+
+def _set_threading_defaults():
+    """Set conservative nested-threading defaults before native libraries load.
+
+    OpenQP parallelizes the native integral and response kernels with OpenMP.
+    BLAS must NOT spawn a second worker-thread layer inside those OMP regions,
+    or it oversubscribes the cores (OMP_threads x BLAS_threads) and stalls --
+    measured 1.53x faster on caffeine/6-31G(d,p) at 24 threads with sequential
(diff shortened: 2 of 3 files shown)


karmachoi:
Addressed the Codex P2 in 5b25e2c3: the macOS `OMP_NUM_THREADS` default is now derived per-host from the performance-core count (`hw.perflevel0.physicalcpu` on Apple Silicon, falling back to `hw.physicalcpu`/`os.cpu_count()`) instead of a fixed 16, so it no longer oversubscribes smaller Apple Silicon or Intel Macs. Still `setdefault`, so an explicit `OMP_NUM_THREADS` wins.

@codex review

PR #192, opened 2026-06-07
input: omp_threads option to set OpenMP threads per rank (build-aware)

karmachoi:
## Summary

Adds a first-class way to set the OpenMP thread count from the OpenQP input (or CLI), instead of relying solely on the `OMP_NUM_THREADS` environment variable, with build-aware handling.

### New: `omp_threads` input option / `--omp` flag

```
[input]
omp_threads = 16      # OpenMP threads per process / MPI rank
```
or on the command line:
```
openqp mol.inp --omp 16
```

**Why a pre-parse.** The OpenMP runtime caches `OMP_NUM_THREADS` when `liboqp` loads, so the value must be set *before* that. The frontend therefore reads `omp_threads`/`--omp` and sets `OMP_NUM_THREADS` **before `import oqp`** (`pyoqp._apply_omp_threads_from_input`). It is registered in the input schema (`OQP_CONFIG_SCHEMA['input']['omp_threads']`, default `0` = leave the environment untouched).

**Precedence** (highest first): `--omp N` → input `omp_threads` → existing `OMP_NUM_THREADS` → built-in default.

### Build-aware

`liboqp` now exports `oqp_have_openmp()` (1 if compiled with OpenMP, else 0; defined in `mathlib/blas_thread_ctl.c`, declared in `include/oqp.h`). If threads are requested but the build has no OpenMP, the frontend prints a clear warning that the run is serial — so the option "adjusts" to the build configuration rather than silently doing nothing.

### MPI note (intentionally not an input keyword)

MPI ranks are spawned by the launcher (`mpirun -np N` / SLURM `srun`) **before** OpenQP runs, so a rank-count input keyword would be misleading. Rank count stays launcher-controlled; the existing `--nompi` flag toggles MPI. The hybrid model is `mpirun -np <ranks>` × `omp_threads = <threads per rank>`.

## Validation

- `omp_threads = 7` in the input → `OMP_NUM_THREADS=7`; `--omp 12` overrides to 12 (precedence).
- `oqp_have_openmp()` returns 1 for the OpenMP build (no false warning); the warning path triggers only for a serial build.
- SCF energies unchanged.

## Compatibility

- Default `omp_threads=0` leaves current behavior (env-driven) exactly as-is.
- No new build dependencies; the C helper is compiled with the existing flags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,27 @@ For OpenMP and MPI run:
 mpirun -np number_of_mpi openqp any_example_file.inp
 ```
 
+##### Setting the OpenMP thread count
+
+The number of OpenMP threads (per process / MPI rank) can be set directly from
+the input file or the command line, instead of relying on `OMP_NUM_THREADS`:
+
+```ini
+[input]
+omp_threads=16        # OpenMP threads per process / MPI rank
+```
+
+```bash
+openqp any_example_file.inp --omp 16
+```
+
+Precedence (highest first): `--omp` → input `omp_threads` → `OMP_NUM_THREADS` →
+built-in default. The value is applied before the OpenMP runtime initializes, and
+is honored on the programmatic `Runner(input_dict=...)` path as well. If OpenQP
+was built without OpenMP (`-DENABLE_OPENMP=OFF`), the request is ignored with a
+warning and the run is serial. MPI rank count remains launcher-controlled
+(`mpirun -np ...`); `--nompi` disables MPI.
+
 ### Detailed Documentation
 
 For more in-depth information, visit:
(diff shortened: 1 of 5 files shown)


karmachoi:
Addressed the Codex P2 in d2ebf81b: `omp_threads` is now honored on the programmatic `Runner(input_dict=...)` path too. Since the pre-import hook can only see a CLI flag or an input *file*, `Runner.__init__` now applies the parsed `omp_threads` at runtime via a new `oqp_omp_set_num_threads()` (`omp_set_num_threads`, exported from `blas_thread_ctl.c`/`oqp.h`) and exports `OMP_NUM_THREADS` for consistency; it warns if the build has no OpenMP. Verified: `Runner(input_dict={'input':{'omp_threads':5,...}})` -> `OMP_NUM_THREADS=5`.

@codex review

PR #200, opened 2026-06-09
fix: default auto TRAH to native implementation

karmachoi:
## Summary
- Make native TRAH the default/first implementation for trh_impl=auto
- Set trh_impl numeric mapping to native=0 and OTR=1
- Remove the DFT XC integration timing line from SCF output

## Verification
- python3 -m py_compile pyoqp/oqp/molecule/oqpdata.py
- git diff --check
- git grep -n "DFT XC integration time" -- source pyoqp returned no matches

### DIFF
--- a/pyoqp/oqp/library/single_point.py
+++ b/pyoqp/oqp/library/single_point.py
@@ -527,15 +527,41 @@ class SinglePoint(Calculator):
         fallback = getattr(self, 'alternative_scf', scf_config.get('alternative_scf', 'trah'))
         stability = getattr(self, 'stability', scf_config.get('stability', False))
         trah_stab_default = scf_config.get('trh_stab', False)
+        trh_impl = str(scf_config.get('trh_impl', 'auto')).strip().lower()
+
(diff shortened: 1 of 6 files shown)


karmachoi:
Updated in 18f84d67: trh_impl=auto now keeps native as the first choice in the SCF manager and falls back to OpenTRAH only if the native TRAH stage does not converge. The numeric mapping remains native=0, auto=0, otr=1, and explicit trh_impl=otr/native still selects only that implementation.

Verification:
- python3 -m py_compile pyoqp/oqp/molecule/oqpdata.py pyoqp/oqp/library/single_point.py
- git diff --check
- git grep -n "DFT XC integration time" -- source pyoqp returned no matches
- Ran ~/clone/oqp-inputs/cheol/scf_problems/fulvene_from_Josene.inp with the rebuilt package: native TRAH was selected; the DFT XC timing line was absent. The run reached TD Z-vector but reported Z-vector not converged.

karmachoi:
Replaced with #201

PR #202, opened 2026-06-10
Fix MRSF-EKT IP/EA Dyson root output

Alireza-Lashkaripour:
Summary:
- Prints/stores all retained MRSF-EKT Dyson roots instead of capping by the parent TDDFT nstate.
- Skips near-zero EKT eigenvalues before JSON output/comparison.
- Refreshes the H2O IP/EA example references and scaffold tests.

Examples:
- H2O IP example now stores 5/13 retained roots.
- H2O EA example now stores 8/13 retained roots.
- C2H4 EA weak-root discrepancy was traced to GAMESS grid settings; fine-grid GAMESS agrees with OpenQP at ~0.001 eV.

Tests:
- `python -m unittest tests.test_mrsf_ekt_scaffold`: 13 tests OK.
- H2O MRSF-EKT IP `run_tests`: Passed 1/1.
- H2O MRSF-EKT EA `run_tests`: Passed 1/1.


karmachoi (commented):
I found one blocker before merge: the EA-only relaxed-density `calc_fock` call appears to leak its temporary energy back into the final molecule results.

In `source/modules/tdhf_mrsf_ekt.F90`, the EA path saves/restores `FOCK_A`/`FOCK_B` after calling `calc_fock`, but `calc_fock` also writes `infos%mol_energy%energy = E%etot` in `source/scf_addons.F90`. Since the final JSON writes `self.mol_energy.energy`, the refreshed EA reference now changes top-level `energy` from the prior SCF value `-76.05983111683781` to `-76.36851222961076`.

Suggested fix: snapshot the relevant `infos%mol_energy` value/state before the EA `calc_fock` call, restore it afterward, and regenerate the EA reference JSON so the top-level energy remains the SCF/reference energy while the EKT EA records use the rebuilt relaxed-density Fock.

karmachoi:
Let's merge it. Great work!

PR #205, opened 2026-06-13
Add MRSF NAMD, SOC-NAMD, and ESPF QM/MM dynamics

karmachoi:
## Summary

This PR adds the OpenQP NAMD/QM-MM development stack for MRSF-TDDFT dynamics:

- native Tully FSSH NAMD (`runtype=namd`) for gas-phase internal conversion
- overlap-derived MRSF NAC/TDC propagation, including NPI/logarithmic TDC support
- SOC-NAMD in the SHARC-like spin-adiabatic representation
- SOC force variants for energy-conservation testing:
  - Option 1: spin-adiabatic weighted MCH gradient
  - Option 2: finite-difference `dU/dt` correction
  - Option 2b: TDC-projected derivative-coupling correction
  - Option 3: MCH-basis SOC propagation with exact active-root gradients
- ESPF QM/MM NAMD with OpenMM, PME support, rigid-water constraints, and ESPF force fixes
- SOC-QMMM production-mode guards and regression coverage
- documentation of the remaining fully analytic spin-adiabatic SOC-gradient target

## Notes

The recommended production path from the validation work is currently the MCH-basis SOC mode (`soc_basis=mch`, Option 3), because it uses exact active-root MCH gradients and avoids the approximate spin-adiabatic weighted-gradient force. The fully analytic spin-adiabatic force remains future work because it requires full MCH derivative-coupling vectors and SOC-gradient matrix elements.

This branch has been merged with current `upstream/main`; the only merge conflict was in `pyoqp/oqp/pyoqp.py`, resolved by keeping the NAMD/QMMM initialization path while preserving upstream build-version logging.

## Validation

- `python -m py_compile pyoqp/oqp/pyoqp.py`
- `python -m unittest tests/test_soc_namd_qmmm_production.py tests/test_rohf_status_and_interface.py`
  - 13 tests run, 1 skipped


iamDongHwan:
I ran into an issue when specifying a PDB file with atom indices in the "system" section, for example:

system = nvt_final.pdb 1 2 3 4 5 6 7 8 9 10

This produces the following error:

[ERROR] input.system: Referenced XYZ file does not exist.
current: 'nvt_final.pdb 1 2 3 4 5 6 7 8 9 10'
fix: Fix the path or place the XYZ file in the working directory.
note: Set system to an XYZ file path or inline coordinates with one atom per indented line.

As a temporary solution, I commented out those lines in /pyoqp/oqp/pyoqp.py.

from oqp.utils.input_checker import check_input_values

_input_file = getattr(self.mol, "input_file", None)
check_input_values(
    self.mol.config,
    input_dir=os.path.dirname(os.path.abspath(_input_file)) if _input_file else None,
)

After commenting out those lines, the calculation works correctly with the PDB input. Maybe we need some corrections of /pyoqp/oqp/utils/input_checker.py

iamDongHwan:
Also, in line 82–83 of "pyoqp/oqp/utils/qmmm.py",

if constraints is None:
   const = None

should be corrected to

if constraints is 'None':
   const = None

to avoid the error below:
"Unknown type of constraint None, select a value between None, HBonds, AllBonds or HAngles"

karmachoi:
Bug Fixed.

> **1. `.pdb` QM/MM system fails input validation ("Referenced XYZ file does not exist")**
> The validator in `input_checker._check_system` was treating the entire first line of `[input] system` as a filename, so a QM/MM spec like `ala.pdb 9 10 17 18 19` was checked as a path literally named `"ala.pdb 9 10 17 18 19"` and always failed — which is why commenting out validation in `pyoqp.py` worked around it.
> Fix: only the first whitespace-delimited token is taken as the path. A `.pdb` path is validated for existence, and the trailing QM-atom indices are validated separately (integers or `start-end` ranges, non-negative, unique) via a new `_check_qm_atom_indices` helper that mirrors the parsing in `oqpdata.read_system`. So `system = mol.pdb 9 10 17-19` now passes validation, and the workaround is no longer needed.
>
> **2. Constraint check uses identity instead of equality ("Unknown type of constraint None")**
> `_openmm_from_oqp` compared `constraints is None`, but the config supplies the **string** `"None"` (the default), so it fell through to the error branch. Fix: `None`, `""`, and `"none"` (case-insensitive) are all treated as "no constraint"; genuinely unknown values still raise as before.
>
> Both are pure-Python paths, so I verified them directly without a full build: valid QM/MM specs pass; missing PDB, missing/duplicate/malformed indices each produce a precise error; and every constraint spelling (`None`/`none`/`""`/`HBonds`/`AllBonds`/`HAngles`) resolves correctly while a bogus value still errors.

karmachoi:
**Title:** `Add MRSF NAMD, SOC-NAMD, and ESPF QM/MM dynamics (with covalent-boundary QM/MM)`

**Body:**

> This PR introduces nonadiabatic molecular dynamics capabilities to OpenQP:
>
> - Native Tully FSSH surface-hopping for gas-phase internal conversion
> - Overlap-derived MRSF NAC/TDC propagation with NPI/logarithmic support
> - SOC-NAMD in the spin-adiabatic representation with multiple force variants
> - ESPF QM/MM NAMD with OpenMM, PME support, and rigid-water constraints
> - **QM/MM boundary (dangling-bond) connectivity:** the `OpenQpQMMM` driver now caps covalent bonds cut by the QM/MM partition with hydrogen link atoms (scaled-position `g = (r_H+r_QM)/(r_QM+r_MM)`), redistributing link-atom gradients onto their QM/MM hosts and folding link charges onto the host. Whole-molecule QM regions (e.g. a water box) are unaffected — the path is a no-op when no bond is cut. New unit-tested module `qmmm_connectivity.py`.
> - **QM/MM input hardening (review fixes):** `.pdb <indices>` system specs no longer fail validation with a spurious "Referenced XYZ file" error, and the `constraints="None"` default is handled by value rather than identity.
> - Production-mode guards and regression coverage
>
> The recommended production approach uses MCH-basis SOC mode with exact active-root gradients, avoiding approximate spin-adiabatic weighted gradients.
>
> **Validation:** `python -m py_compile` across the Python entry points; unit tests pass (`tests/test_qmmm_connectivity.py` added — 8 cases covering detection, geometry, and gradient conservation). Link-atom energy/force consistency additionally checked end-to-end by finite differences on the alanine QM/MM example.

karmachoi:
## Merged `main` + resolved conflicts (branch now mergeable)

Merged current `main` into `codex/soc-namd-options` and resolved all conflicts; GitHub now reports the PR as **MERGEABLE**.

### Conflict resolutions (7 files)
- **`source/types.F90`, `include/oqp.h`** — unioned the `control_parameters` struct: kept `main`'s perf knobs (`xc_c2f`, `xc_phi_cache`, `xc_incdft`, `grad_cutoff`, `mrsf_resp_cutoff`, `mrsf_fp32`, `mrsf_zv_warmstart`) and this branch's `qmmm_flag`, in the **same field order** in both files (cffi reads `oqp.h`; the Fortran `bind(c)` type must match).
- **`source/modules/int1e.F90`** — took `main`'s `alloc_or_die` allocation (allocates-or-reuses in place, so the Python-set `OQP::mm_potential` tagarray is preserved); dropped the legacy `remove_records`.
- **`source/scf.F90`** — kept both the QM/MM `hcore_bk` backup and `main`'s IncDFT reset.
- **`pyoqp/oqp/openqp.py`** — took `main`'s rewritten module and reconciled the legacy `OPENQP` wrapper the ESPF QM/MM driver depends on: it now accepts an optional `silent` arg (`OPENQP(cfg, True)`) and exposes `self.sp` (`SinglePoint`) for the driver's embedded SCF/excitation calls.
- **`README.md`** — adopted `main`'s capability tables; re-added a "Dynamics & QM/MM" subsection.
- **`.github/workflows/CI.yml`** — took `main`'s example-validation / feature-coverage / test-suite gates.

### Build fix required by the merge
`main`'s tagarray refactor removed `container_t%remove_records` and `%reserve_data` (now `%erase` / `%alloc_or_die`). `source/modules/namd.F90` and `source/modules/qmmm.F90` still used the old API and failed to compile after merging:
```
Error: 'remove_records' is not a member of the 'container_t' structure
Error: 'reserve_data'   is not a member of the 'container_t' structure
```
Migrated those calls to `alloc_or_die`/`erase` (matching how the rest of the tree already uses them). **The full tree now builds** (`liboqp.dylib` + the cffi `_oqp` extension, gcc-15/gfortran-15, Release, netlib ILP64) and a plain MRSF energy runs correctly end-to-end (H2O BHHLYP: SCF −76.07744682, MRSF converged in 3 iters).

### Codex review findings — all already resolved on the current head
Re-verified each against the code; no change needed:
- **P1 `qmmm_driver.py` (MM-MM exceptions)** — the exception loop now requires ≥1 QM atom (`int(p1) in qm_set or int(p2) in qm_set`), so MM-MM exceptions are preserved.
- **P1 `pyoqp.py` (QM/MM energy dispatch)** — dispatch to `QMMM_MD` is now gated on `runtype == 'md'` only; energy/grad/namd stay on the Runner path.
- **P2 `namd.py` (trivial-crossing active state)** — both FSSH loops apply the returned state via `active_changed = new_active != active_old`.
- **P2 `tdhf_mrsf_z_vector.F90` (ESPF gate)** — the ESPF fit is gated on `infos%control%qmmm_flag`.

### Pre-existing issues found during verification (not introduced by this merge)
These reproduce on the pre-merge head and are worth a follow-up:
1. **Single-point QM/MM energy (`qmmm_flag=true`, `runtype=energy`) aborts** with `qmmm_espf::add_potqm_contributions: Record 'OQP::ESPF_CORR' not found`. The SCF calls `add_potqm_contributions` whenever `qmmm_flag` is set, but `ESPF_CORR` is only created by `espf_op_corr`, which the `single_point.py` path never calls (it is commented out at ~L817/L836 — only `qmmm_driver.py` / `namd.py` call it). So the `examples/QMMM/*_energy.inp` decks don't run through the Runner path.
2. **Stale basis name in the QMMM examples** — `examples/QMMM/ala.inp` and `2E4E_*.inp` use `basis=6-31gs`, which the current basis library rejects (`6-31g*` works).

PR #206, opened 2026-06-13
[codex] Cache Docker build dependencies

karmachoi:
## Summary

This PR reduces repeated work in the Docker image CI without changing the normal local build path documented in the README.

- Use Docker Buildx with GitHub layer cache for the image build.
- Persist BuildKit cache mounts for pip and OpenQP bundled externals in GitHub Actions.
- Install Python build/runtime dependencies in a source-independent Docker layer derived from `pyproject.toml`.
- Run the final source install with `--no-build-isolation --no-deps` so source-only PRs do not repeatedly resolve/download Python dependencies.
- Point Docker-only CMake builds at `/root/.cache/openqp/externals` so bundled externals can be reused inside Docker CI.

## Cache invalidation

- `pyproject.toml` changes invalidate the Python dependency layer and the mount-cache key, then the refreshed cache is saved for later builds.
- `external/CMakeLists.txt` changes invalidate the Docker mount-cache key, while OpenQP's internal externals key continues to separate incompatible external builds by toolchain, flags, BLAS/LAPACK settings, build type, and bundled external versions.
- Source-only changes still rebuild OpenQP itself, but should reuse the Python dependency setup and compatible bundled externals.

## Validation

- `git diff --check`
- Parsed `.github/workflows/docker-build.yml` with Ruby YAML loader
- `docker buildx build --check .` completed Dockerfile frontend checks locally; it reported only the expected platform warning because this Mac is arm64 while `openqp/openqp-buildenv:1` is linux/amd64.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -20,6 +20,10 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v4
 
+      - name: Set up Docker Buildx
+        id: setup-buildx
+        uses: docker/setup-buildx-action@v3
+
       - name: Log in to Docker Hub
         if: github.event_name == 'push' && env.HAS_DOCKER_CREDENTIALS == 'true'
         env:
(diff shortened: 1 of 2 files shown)


karmachoi (commented):
## Approve — clean, low-risk CI caching optimization

CI-only change (2 files, +63/-3); the documented local `pip install .` path is untouched. The Docker `build` job (runs on `pull_request → main`, ~18m) exercises the entire new path end-to-end and passes, along with the full gcc-build matrix.

**Verified the things that could plausibly have broken it, but don't:**
- `import tomllib` is safe — buildenv is `ubuntu:24.04` (Python 3.12, ≥3.11).
- `-DOQP_EXTERNALS_ROOT` binds to existing infra already on `main` (`CMakeLists.txt` / `external/CMakeLists.txt`), not something added here.
- `--no-deps` is sound: the source-independent layer installs `project.dependencies` *with* their transitive deps (no `--no-deps` there), and `--no-build-isolation` is satisfied because `build-system.requires` is installed in that same layer.
- String-only dedup leaving `numpy` + `numpy>=1.20.0` (and `cffi` + `cffi>=1.16.0`) is harmless — both constraints resolve together in one `pip install`.
- `buildkit-cache-dance` extract runs as a post-action *before* `actions/cache` save (reverse declaration order), so the save captures the freshly-updated mount caches.

**Minor, non-blocking:** branch is 2 commits behind `main` (still mergeable) — a rebase before merge would let CI reflect current main. The workflow's "Multi-Platform" name remains single-platform, but that's unchanged from the prior `docker build` (no regression).

PR #207, opened 2026-06-13
full-density source projection for ddPCM SCF energies

Alireza-Lashkaripour:
## Summary
Adds an **energy-only ddPCM reaction field for the RHF/ROHF reference SCF** via the ddX library, using the **full self-consistent density** as the solute source. Scope is the reference-SCF single-point energy; gradients and state-specific MRSF-PCM are out of scope (planned).

## What existed before
The initial ddX/ddPCM path built the solute source handed to the solver from **atom-centered Mulliken multipoles truncated at quadrupole (ℓ≤2)**. Both the cavity potential and the adjoint source Ψ were formed from those few moments.

## The problem
The cavity surface lies only ~1 vdW radius from the nuclei — inside the region where a low-order multipole expansion of an extended density is not yet asymptotic. For anisotropic densities (multiple bonds, lone pairs) the ℓ≤2 source misrepresents the potential exactly where the apparent surface charge is built, giving errors up to **~4 kcal/mol** versus a full-density reference. (Separately, a factor `f(ε)=(ε−1)/ε` was applied on top of ddX's already-dielectric ddPCM amplitude, **double-counting** the screening — invisible to a variational SCF and to a reference carrying the same factor.)

## The fix
- Project the **full AO density** onto a Becke-partitioned molecular grid (Treutler–Ahlrichs sizing) to build the exact cavity potential φ and the full-density adjoint Ψ in ddX's harmonic convention (ℓ≤8), with an exterior-leak continuation.
- Reaction-field energy `ΔG = −½⟨φ, q⟩`, **no extra dielectric factor** (ddX's R_ε already carries it — verified to machine precision against the analytic Born ion). RHF/ROHF Fock coupling `V_pcm = −ext(q)`; for ROHF the spin-free V enters both blocks.
- Removed the `f(ε)` double-count.
- Gated the convention self-diagnostics behind `verbose≥3`: production does **one** ddX solve and emits no diagnostic log lines.
- Adds `examples/PCM/` (RHF spherical-harmonic + open-shell ROHF) with reference JSON.

## Validation vs PySCF ddCOSMO
Matched protocol (RHF/6-31G*, water ε=78.3553, ℓmax=8, 302-point Lebedev, η=0.1, Bondi radii); only the solute source differs between the two OpenQP columns. Δ = OpenQP − PySCF (kcal/mol).

| Molecule | Baseline (ℓ≤2) | This work (full-ρ) | PySCF ddCOSMO | Δ baseline | Δ this work |
|---|---:|---:|---:|---:|---:|
| H2O | -0.017456 | -0.018755 | -0.018843 | +0.870 | +0.055 |
| NH3 | -0.015040 | -0.015134 | -0.015173 | +0.084 | +0.025 |
| HF | -0.012499 | -0.014153 | -0.014256 | +1.102 | +0.064 |
| CH4 | -0.000972 | -0.000499 | -0.000508 | -0.291 | +0.005 |
| CO | -0.003092 | -0.003325 | -0.003309 | +0.136 | -0.009 |
| CO2 | -0.010077 | -0.011375 | -0.011388 | +0.822 | +0.008 |
| HCN | -0.015314 | -0.021477 | -0.021540 | +3.907 | +0.040 |
| H2CO | -0.013534 | -0.019003 | -0.019063 | +3.470 | +0.038 |
| CH3OH | -0.018533 | -0.015378 | -0.015508 | -1.898 | +0.081 |
| CH3CN | -0.016806 | -0.023419 | -0.023390 | +4.131 | -0.018 |
| **MAE** | | | | **1.671** | **0.034** |
| **Max** | | | | **4.131** | **0.081** |

The full-density residual (**MAE 0.034**) sits at the ddPCM-vs-ddCOSMO model floor; the baseline's **1.67 kcal/mol** is the ℓ≤2 truncation. Test suite: `Ran 100 tests, OK (skipped=41)`.

## Notes
- Built with `-DENABLE_DDX=ON`; ddX is a dynamically-linked LGPL-3.0 backend.
- `source/solvent_pcm_harmonics.F90` replicates ddX's harmonic routines to guarantee the multipole convention matches the linked library — flagging for an attribution-header / licensing decision.


Alireza-Lashkaripour:
### DFT + PCM cross-code benchmark (follow-up to the HF table)

**PBE**

| Molecule | OpenQP | PySCF | Δ (kcal/mol) |
|---|---:|---:|---:|
| H2O | -0.016598 | -0.016657 | +0.037 |
| NH3 | -0.014214 | -0.014226 | +0.007 |
| HF | -0.012058 | -0.012154 | +0.061 |
| CH4 | -0.000619 | -0.000632 | +0.008 |
| CO | -0.002129 | -0.002132 | +0.002 |
| CO2 | -0.006153 | -0.006093 | -0.038 |
| HCN | -0.015728 | -0.015737 | +0.006 |
| H2CO | -0.010720 | -0.010713 | -0.004 |
| CH3OH | -0.012719 | -0.012804 | +0.053 |
| CH3CN | -0.019232 | -0.019148 | -0.053 |
| **MAE / Max** | | | **0.027 / 0.061** |

**BHHLYP** (50% exact exchange)

| Molecule | OpenQP | PySCF | Δ (kcal/mol) |
|---|---:|---:|---:|
| H2O | -0.017790 | -0.017854 | +0.041 |
| NH3 | -0.014662 | -0.014680 | +0.011 |
| HF | -0.013248 | -0.013341 | +0.058 |
| CH4 | -0.000547 | -0.000557 | +0.006 |
| CO | -0.002505 | -0.002494 | -0.007 |
| CO2 | -0.008792 | -0.008748 | -0.028 |
| HCN | -0.018595 | -0.018642 | +0.029 |
| H2CO | -0.014895 | -0.014912 | +0.010 |
| CH3OH | -0.014216 | -0.014316 | +0.063 |
| CH3CN | -0.021144 | -0.021098 | -0.029 |
| **MAE / Max** | | | **0.028 / 0.063** |

source/solvent_pcm_harmonics.F90
```diff
@@ -0,0 +1,422 @@
...
+!> adjoint source via
+!>
+!>     psi(lm,isph) = 4*pi/((2l+1) * rsph(isph)^l) * M_lm(isph)            (ddX)
+!>
+!> (see ddx_multipolar_solutes.f90::multipole_psi). For a QM solute the
+!> physically correct full-density source therefore requires the SAME multipole
+!> convention ddX uses internally; otherwise the per-l normalisation/ordering
+!> silently disagrees and Psi is wrong.
+!>
+!> To guarantee an exact match this module replicates, verbatim (only constants
```
karmachoi: This file says it replicates ddX harmonic routines verbatim. Since ddX is LGPL-3.0 and OpenQP is GPLv3/commercial dual-license, please add an explicit attribution/license header or choose a non-vendored approach before merge. The current CMake comments carefully keep ddX external, so copying implementation code into `source/` should be a deliberate licensing decision.

karmachoi (commented):
Thanks for the heavy lift here. The implementation direction looks coherent to me, but I would like two things clarified before merge:

1. The ddX harmonic routines copied into OpenQP need an explicit licensing/attribution decision. ddX is LGPL-3.0 and OpenQP is GPLv3/commercial dual-license, so this should be handled deliberately rather than left as an implicit source copy.

2. The visible GitHub CI checks build/install the project, but I do not see a required check running ctest or the ddX-enabled PCM Python gates. The local source-level tests are useful, but the real ddX SCF benchmark rows skip when no ddX build is present. Please either add a required validation job for the ddX path or document the exact ddX-enabled validation command/build artifact expected for reviewers.

Also, the DFT benchmark tables in the PR comment are useful context, but the committed benchmark data still keeps those DFT rows deferred/non-gating. That is fine for the stated RHF/ROHF HF-first scope, but the PR text should keep that distinction explicit unless those rows are promoted to verified regression gates.

karmachoi (approved):
Great!. It works well.

PR #211, opened 2026-06-17
fix(int1): add barrier to one-electron integral worksharing loops nested in replicated outer loops

karmachoi:
## Summary

Ten one-electron integral routines in `source/integrals/int1.F90` share a parallel pattern that can desynchronize OpenMP worker threads. Each has an `!$omp do schedule(dynamic)` over the **inner** shell index nested inside an **outer** shell loop that *every thread executes* (the outer `DO` is not a worksharing construct). The inner loop ended with `!$omp end do nowait`.

Without the implicit barrier, a thread that finishes its chunk of the inner loop races ahead into the next outer iteration and re-enters the worksharing region while other threads are still finishing the current one. With `schedule(dynamic)` this leaves threads spread across two live instances of the same worksharing construct, corrupting the work-share dispatch state. This was observed as an **intermittent segfault in `basis_overlap`** under OpenMP, and is a latent correctness/stability hazard in the siblings.

Removing `nowait` restores the per-outer-iteration barrier, so all threads finish the inner shell loop before any advances to the next outer shell.

## Why this is safe / cheap

- Adding a barrier can never produce incorrect results — worst case is a small synchronization cost.
- These are **one-electron** integral routines (overlap, kinetic, multipole, external-charge, GIAO), evaluated **once per geometry**, not per SCF iteration. The ERI/Fock hot path is untouched. So the added sync cost is negligible.

## Scope

**Fixed (same nested pattern):**
`basis_overlap`, `overlap`, `kin_ovl_ints`, `mult_all_ints`, `mult_ints`, `int1_coul_ext_chg`, `int1_coul_ext_chg_ewaldsr`, `amom_ints`, `giao_overlap_deriv_ints`, `giao_h10_core_ints` — 10 sites.

**Deliberately left unchanged:**
`lzints` and the second parallel region of `int1_coul_ext_chg_ewaldsr` — there the `!$omp do` is a single top-level worksharing loop (not nested in a replicated outer loop), so `nowait` is harmless before the parallel region's own closing barrier.

## Diff

10 single-line changes, each `!$omp end do nowait` → `!$omp end do`. No logic changes.

### DIFF
--- a/source/integrals/int1.F90
+++ b/source/integrals/int1.F90
@@ -973,7 +973,7 @@ contains
             CALL update_rectangular_matrix(shi, shj, sblk, s)
 
         END DO
-!$omp end do nowait
+!$omp end do
     END DO
 !$omp end parallel
 !   End of shell loops
@@ -1044,7 +1044,7 @@ contains
             CALL update_triang_matrix(shi, shj, sblk, s)
 
         END DO
-!$omp end do nowait
+!$omp end do
     END DO
 !$omp end parallel
 !   End of shell loops
@@ -1122,7 +1122,7 @@ contains
             CALL update_triang_matrix(shi, shj, tblk, t)
 
         END DO
-!$omp end do nowait
+!$omp end do
     END DO
 !$omp end parallel
 !   End of shell loops
@@ -1184,7 +1184,7 @@ contains
             end do
 
         END DO
-!$omp end do nowait
+!$omp end do
     END DO
 !$omp end parallel
 !   End of shell loops
@@ -1246,7 +1246,7 @@ contains
             end do
 
         END DO
-!$omp end do nowait
+!$omp end do
     END DO
 !$omp end parallel
 !   End of shell loops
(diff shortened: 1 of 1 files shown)


karmachoi:
Build-validated: with all 10 edits applied, `int1.F90` compiles cleanly (incremental Fortran build, gfortran-15 + OpenMP, no warnings/errors on the changed file). The edits are a mechanical removal of the optional `nowait` clause, yielding the canonical `!$omp end do`, so there is no interface or logic change.

karmachoi:
Verified — this is correct and resolves the crash. The `basis_overlap` change is exactly the one I confirmed empirically: it turned a segfaulting solvated QM/MM PME NVE into a clean 100 fs (201-step) run, and the other nine sites are the identical barrier-only transform (can only add synchronization, never change results).

One small nit in the PR description, not the code: the **"deliberately left unchanged"** list names `lzints` and the second `ewaldsr` region, but the two `!$omp end do nowait` that actually remain in `int1.F90` after this PR are in **`nuc_ints`** (~L1374) and **`int1_el_pot`** (~L1609). Both are correctly left untouched — each is a *top-level* `!$omp do` immediately followed by `!$omp end parallel`, so the region's own closing barrier already covers it (no replicated outer `DO` wraps them). Might be worth updating the description to name those two so a future reader can verify the scope.

PR #216, opened 2026-06-21
Add automated release packaging workflow

karmachoi:
## Summary
- Replace the old Linux-only wheel job with a release-oriented workflow that builds sdist, Linux x86_64 wheels, macOS x86_64 wheels, and macOS arm64 wheels.
- Publish artifacts to GitHub Releases and PyPI only from the GitHub Release published event.
- Add release packaging notes for the openqp install path, PyPI Trusted Publishing, and the separate install-site repo.

## Validation
- Parsed the workflow YAML locally and verified triggers/jobs.
- Ran git diff --check.
- Ran the package-local runtime-root resolver tests relevant to wheel layout.

## Follow-up
- Configure PyPI Trusted Publisher for the openqp project with repository Open-Quantum-Platform/openqp, workflow .github/workflows/build_wheels.yml, environment pypi.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -8,6 +8,7 @@ name: Build and publish wheels
       - "CMakeLists.txt"
       - "MANIFEST.in"
       - "pyproject.toml"
+      - "basis_sets/**"
       - "cmake/**"
       - "external/**"
       - "include/**"
@@ -108,7 +109,7 @@ jobs:
         env:
           CIBW_PLATFORM: ${{ matrix.platform }}
           CIBW_ARCHS: ${{ matrix.archs }}
-          CIBW_BUILD: "cp39-* cp310-* cp311-* cp312-* cp313-*"
+          CIBW_BUILD: "cp39-* cp310-* cp311-* cp312-* cp313-* cp314-*"
           CIBW_SKIP: "*-musllinux_* pp*"
           CIBW_BEFORE_ALL: ${{ matrix.before_all }}
           CIBW_ENVIRONMENT: ${{ matrix.environment }}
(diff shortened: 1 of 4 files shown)


karmachoi:
Addressed the two packaging review comments in 4cb4d4e:\n\n- Added basis_sets/** to the pull_request path filter so basis/data-only packaging changes run the wheel/sdist workflow.\n- Added cp314-* to CIBW_BUILD so Python 3.14 wheels are included in release builds.\n\nLocal validation: workflow YAML parsed, git diff --check passed, and the package-local runtime-root resolver tests still pass.

karmachoi:
Investigated the macOS arm64 wheel failure. The wheel built successfully, but delocate failed because _oqp.abi3.so depended on @rpath/liboqp.dylib and had no usable rpath inside the wheel.\n\nFix pushed in fd81c5b: on Apple builds, the packaged CFFI extension now rewrites that dependency to @loader_path/liboqp.dylib, matching the installed wheel layout where _oqp.abi3.so and liboqp.dylib both live under oqp/lib. Added a source-level regression assertion for this packaging guarantee.\n\nLocal validation: workflow YAML check, git diff --check, and targeted runtime-root/package-layout tests passed.

karmachoi:
Follow-up pushed in 2802dbb after the latest wheel validation exposed a separate macOS repair issue. The previous @rpath fix worked, but delocate then rejected the macOS wheels because the bundled Homebrew GCC runtime libraries on macOS 15 require MACOSX_DEPLOYMENT_TARGET=15.0.\n\nThis commit sets that deployment target for both macOS wheel jobs, documents that the first macOS wheels target macOS 15+, and adds a source-level regression assertion for the workflow setting.\n\nLocal validation: workflow YAML parsed, git diff --check passed, and the focused runtime/package-layout tests passed. Fresh wheel validation is running on the updated commit.

PR #217, opened 2026-06-24
Limit PR wheel builds to cached Linux smoke

karmachoi:
## Summary\n- keep ordinary PR release-packaging checks to the source distribution plus one Linux CPython 3.11 smoke wheel\n- restore/cache bundled externals for wheel builds via OQP_EXTERNALS_ROOT so source changes do not rebuild third-party libraries every run\n- run the full Linux/macOS wheel matrix only for PRs labeled release, manual runs, tag pushes, and GitHub Releases\n\n## Notes\n- Added a release label as the opt-in switch for release-preparation PRs.\n- Publishing remains restricted to the GitHub Release event.\n- This only changes GitHub Actions/docs/tests; it does not affect local builds after git clone.\n\n## Validation\n- workflow YAML parsed\n- git diff --check\n- python3 -m unittest tests.test_runtime_root_resolution.RuntimeRootResolutionTests.test_macos_wheel_target_matches_runner_gcc_runtime tests.test_runtime_root_resolution.RuntimeRootResolutionTests.test_pull_requests_use_cached_smoke_wheel_not_full_matrix

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -15,9 +15,6 @@ name: Build and publish wheels
       - "include/**"
       - "pyoqp/**"
       - "source/**"
-  push:
-    tags:
-      - "v*"
   release:
     types: [published]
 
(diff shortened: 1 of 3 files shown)


karmachoi:
Adjusted the externals-cache fix so the workflow no longer sets OQP_EXTERNALS_ROOT. Linux cibuildwheel jobs now set the standard XDG_CACHE_HOME to a cached checkout-local directory, letting OpenQP's existing auto-discovery select /openqp/externals. macOS jobs use the default ~/Library/Caches/openqp/externals path. Added a guard test to keep OQP_EXTERNALS_ROOT out of the wheel workflow.

PR #228, opened 2026-06-25
Fix libecpint MinGW M_PI build

karmachoi:
## Summary

- patch bundled libecpint v1.0.7 so MinGW builds define `M_PI` when `<cmath>` does not expose it
- apply the patch from the libecpint ExternalProject before configuring/building it
- bump the bundled externals cache revision so Windows CI does not reuse the already-extracted unpatched libecpint source
- add a regression assertion for the libecpint patch wiring

## Root cause

The Windows MSYS2 preflight on `main` failed in the package build while compiling libecpint:

```text
src/lib/gaussquad.cpp:86:29: error: 'M_PI' was not declared in this scope
```

MinGW does not guarantee `M_PI` from `<cmath>` unless the source defines it or requests nonstandard math constants. The small patch keeps the upstream source behavior unchanged on platforms that already provide `M_PI`.

## Validation

- `python -m unittest tests.test_opentrustregion_linalg_config tests.test_windows_msys2_packaging`

### DIFF
--- a/cmake/patches/libecpint-v1.0.7-mingw-m-pi.patch
+++ b/cmake/patches/libecpint-v1.0.7-mingw-m-pi.patch
@@ -1,0 +1,15 @@
+diff --git a/src/lib/gaussquad.cpp b/src/lib/gaussquad.cpp
+index e0484eb..5389dad 100644
+--- a/src/lib/gaussquad.cpp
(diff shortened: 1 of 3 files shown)


karmachoi:
Closing because Windows/MSYS2 MSI support is being removed from this release path.

PR #230, opened 2026-06-26
[codex] Add OpenQP-native Python scripting API

karmachoi:
## Summary

- Add an additive `OpenQP` high-level Python wrapper, plus `OQP` as a short alias.
- Make the primary style OpenQP-native and compact:
  - `job.molecule(...)` for geometry and charge; HF/DFT may set multiplicity here.
  - `job.theory(...)` for HF, DFT, and MRSF-TDDFT method setup.
  - `job.workflow.<name>(...)` for scientific workflows that select `runtype`.
  - `job.control(omp_threads=..., usempi=...)` for runtime controls.
  - `job.run()` for execution.
- Keep MRSF-TDDFT multiplicity implicit in the high-level API: `job.theory("mrsf-tddft", ...)` supplies the usual ROHF triplet reference, while raw input files still expose `[scf] multiplicity=3`.
- Rename the preferred gradient helper argument to `state=...`; legacy `grad=...` still works and maps to `[properties] grad`.
- Keep plain energy calculations simple: no workflow call is required for the default energy runtype.
- Add workflow helpers for gradient, Hessian, optimization/crossing searches, NAC/NACME, MRSF-EKT, SOC, PCM/ddX, and NMR shielding.
- Encode current OpenQP capability blockers in the Python helpers:
  - SOC requires MRSF-TDDFT.
  - EKT and NAC/NACME require MRSF-TDDFT.
  - PCM/ddX requires HF/DFT reference-SCF, RHF/ROHF, `backend="ddx"`, `mode="reference_scf"`, and energy-only scope.
  - NMR requires HF/DFT reference-SCF; CGO is RHF-only, while GIAO is the open-shell route.
- Add named geometry setup through `job.molecule(geometry="water", ...)` and `get_geometry(...)`, with reusable runtime helpers and optional PubChem PUG-REST lookup.
- Keep PySCF interoperability as an explicit conversion bridge through `OpenQP.from_pyscf(...)`, not as the main input style.
- Preserve the existing `OPENQP({...})` dotted-key wrapper and `Runner` flow for existing scripts.
- Add focused tests in `tests/test_openqp_api.py` for the workflow/control split, capability blockers, optimizer routing, SOC setup, named geometry resolution, PubChem SDF parsing, PySCF conversion, gradient state aliasing, runtime control, and legacy wrapper behavior.

## Validation

- `python3 -m unittest tests.test_openqp_api`
- `git diff --check`

## Manual

- Companion manual updates were pushed to `Open-Quantum-Platform/openqp-docs@f4bc642`.
- Manual validation: `/tmp/openqp-docs-venv/bin/python -m mkdocs build --strict`.
- Manual now documents implicit MRSF multiplicity, `workflow.gradient(state=...)`, `job.control(usempi=...)`, SOC vs scalar relativistic DKH correction, and visible workflow-section titles.


pyoqp/oqp/openqp.py
```diff
@@ -1,8 +1,273 @@
...
+            updates["input.basis"] = basis
+        if charge is not None:
+            updates["input.charge"] = charge
+        for option, value in kwargs.items():
+            updates[f"input.{option}"] = value
+        return self.set(**updates)
+
+    def hf(self, reference="rhf", runtype="energy", multiplicity=None, **scf_keywords):
+        """Use a compact OpenQP HF setup for ordinary single-reference jobs."""
+        self.input(method="hf", runtype=runtype)
```
karmachoi: Confirmed bug. `set_dft_functional` switches to DFT whenever `functional != ''`, so reusing the object after a DFT setup (e.g. `job.dft("pbe").hf()`) leaves `input.functional` populated and `.hf()` silently runs DFT. `.hf()` should clear `functional` (set it to '') so the HF path is honored.

pyoqp/oqp/utils/geometry.py
```diff
@@ -0,0 +1,201 @@
...
+
+import os
+from urllib.error import HTTPError, URLError
+from urllib.parse import quote
+from urllib.request import urlopen
+
+from oqp.utils.constants import ANGSTROM_TO_BOHR
+
+
+_BOHR_UNITS = {"bohr", "au", "a.u.", "atomic_unit", "atomic_units"}
```
karmachoi: Confirmed bug. PySCF accepts `unit='B'`/`'b'` as Bohr, but those aliases are missing here, so `_is_bohr` returns false for `mol.unit='B'`. The Bohr coordinates are then passed through as Angstrom and OpenQP's reader converts them again, scaling all distances by ~1/0.529. Add `'B'`/`'b'` (and the spelled-out `'bohr'`/`'au'` if not already covered) to the Bohr alias set.

PR #236, opened 2026-06-27
perf(MRSF): faster MRSF-TDDFT response Fock digestion (~16%, + optional FP32)

karmachoi:
## Summary
Speeds up the MRSF-TDDFT linear-response (Davidson sigma) build by optimizing the
Fock-like digestion kernel `int2_mrsf_data_t_update` in `source/tdhf_mrsf_lib.F90`.
Profiling the native CPU MRSF response shows the per-iteration `int2` digestion is
**96–98% of the cost** (density build, back-projection, and the subspace solve are each
<2%), and the kernel is memory-bandwidth bound. Two changes target that directly.

### Commit 1 — exact, default (bit-identical results)
- **Coulomb 8→4**: the 8 permutational read-modify-writes hit only 4 distinct Fock
  targets; collapse them and read a **precomputed symmetric Coulomb density** `ds = d3 + d3ᵀ`
  (built once per run) instead of summing two scattered reads per integral.
- **Array-sections → fused stride-1 loops** (vectorize, no per-statement temporaries).

### Commit 2 — optional FP32 accumulation (`OQP_MRSF_FP32=1`, default off)
- Accumulate the Fock-like build in single precision (FP32 operand copies + an FP32
  per-thread accumulator, folded back to FP64 in `parallel_stop`). Default off = exact FP64.

## Measured (BHHLYP, ROHF triplet ref, cc-pVDZ)
| system | int2 baseline | exact (commit 1) | +FP32 (commit 2) |
|---|---|---|---|
| benzene | 3.21 s | 2.70 s (**~16%**) | 2.51 s (**~22%**) |
| thymine | 17.7 s | 14.9 s (**~16%**) | 13.5 s (**~24%**) |

- **Accuracy**: exact path reproduces excitation energies bit-for-bit; FP32 path within
  ~3–6 µeV with **unchanged Davidson iteration count**.
- int2 dominance grows with system size, so the relative speedup is conservative here.

## Notes
- No public API change; the FP32 mode is opt-in via env var and inert by default.
- For context, several other directions were profiled and found not to help on CPU and are
  **not** included: improved preconditioner (Olsen) and warm-start (MRSF is already near its
  Davidson iteration floor), and Schwarz-routed mixed precision (the digested integrals are
  already screened, so routing yields no net gain over plain FP64/FP32).


karmachoi:
## Companion note: XC-grid reuse for the response build (composes with this digestion + FP32 work)

A literature scan on faster numerical XC integration turned up a reuse pattern that stacks cleanly on top of this PR's Fock-digestion speedup. Sharing it here as input for the next step. Numbers/claims are cited.

### The invariant
For fixed geometry+basis the integration grid and the **collocation matrix `Φ_{μi}=φ_μ(r_i)` depend only on geometry, never on the density** — and the response build adopts the ground-state functional + quadrature grid by construction. So each Davidson trial-vector build is structurally a ground-state Vxc build with the kernel `f_xc = d²E_xc/dD²` (evaluated at the frozen ground-state density) swapped in for `v_xc`. That makes the XC grid work in the response solve almost entirely cacheable.

### What to cache once — two tiers
**Tier A — geometry-only** (cache at grid setup; valid for the whole SCF *and* the whole Davidson solve):
- collocation `Φ`, `∇Φ` (+2nd derivs for meta-GGA gradients) per batch
- atomically-scaled SSF weights `w` + the zero-weight mask
- per-batch significant-shell lists + batch→grid maps

**Tier B — SCF-density-only** (cache after SCF converges; constant across the entire response solve):
- ground-state `ρ, ∇ρ, τ`, the `v_xc`, and the second derivatives `f_xc` (`∂²ε/∂ρ², ∂²ε/∂ρ∂γ, ∂²ε/∂γ²`, + τ terms); third derivatives `g_xc/k_xc` for excited-state gradients

With A+B cached, the **only** per-Davidson-iteration grid cost is: (1) two collocation GEMMs `δρ_j = Φ_jᵀ T_j Φ_j`, (2) a pointwise multiply of `δρ` by the cached `f_xc`, (3) a SYR2K back to AO space `F^x_j = Z_j Φ_jᵀ + Φ_j Z_jᵀ`. No re-evaluation of Φ, weights, ρ, or functional derivatives. (GPU4PySCF-ris does exactly this — reuses ground-state density/gradients/Hessians on the same grid; Pu et al., arXiv:2511.18233.)

### Why it fits MRSF and this PR
- The kernel **contraction is the bottleneck**, not grid evaluation — so batch it over (trial vectors × roots) to fill BLAS, which is the same data-layout/loop-reorder lever that already gave ~4× in the GAMESS MRSF build (Lee, Filatov, Choi et al., *J. Chem. Phys.* 158, 194105 (2023)). The whole Davidson stage there is ~0.23× the SCF wall time; cached grid reuse keeps it firmly sub-SCF.
- MRSF's collinear kernel leaves only exact exchange in the spin-flip block, so caching the exact-exchange grid quantities accelerates `g_xc`. Seminumerical exact-K (COSX/sn-K) for that block runs on the **same** cached `Φ`+weights (half-transformed integrals + overlap-fit `Q` are trial-vector-independent) — ~2–4× on excitation energies at 50–150 atoms, 9.4× on a 585-bf cluster (Neese, *Chem. Phys.* 356, 98 (2009); Helmich-Paris et al., *JCP* 155, 104109 (2021)).
- Noncollinear SF kernels `f_xc^SF ~ (v_xc^α − v_xc^β)/(ρ_α − ρ_β)` are singular where `ρ_α ≈ ρ_β`; stabilize pointwise on the cached ground-state `ρ^{α/β}, v_xc^{α/β}` via `max(|ρ_α−ρ_β|, T)` with `T≈1e-4` (Hernández-Segura & Luber, *JPCA* 2025) — precomputed once per SCF density, reused every response iteration.

### On the FP32 option in this PR
This aligns with the literature consensus: spend reduced precision on the **ERI/exchange digestion** (as done here), but keep the **numerical XC grid sum in FP64** — FP32 on the grid sum gives ~1e-5 a.u. errors that perturb the kernel (GauXC/NWChemEx deliberately keep XC in double). TeraChem-style mixed precision (FP64 above a Schwarz bound ~1e-5, FP32 below) ~halves cost with CIS/TDDFT excitations within ~0.0005 eV — i.e. the FP32 win lives on the integral side, which is exactly where this PR puts it.

### Suggested order (highest payoff-per-effort)
1. Cache Tier A + Tier B at end of SCF and feed every Davidson build (storage/plumbing; the grid invariant already holds).
2. Restructure the response XC-kernel contraction into batched GEMM/SYR2K over (trial vectors × roots) — reuses the SCF Vxc skeleton with `f_xc` for `v_xc`.
3. If not already present in the ground-state XC core: SSF compact-support weights + significant-shell screening + spatial batching, and SG-2/SG-3 as the default grid (SG-3 for any Minnesota functional; never SG-1 there). This metadata *is* the Tier-A cache, so it pulls double duty.

Through-line: the grid is the invariant — cache the geometry-only and SCF-density-only artifacts once, reuse them across every response iteration, and do the contraction in batched BLAS. This composes with (doesn't overlap) the Fock-digestion + FP32 work in this PR.

<sub>Compiled from a cited literature review (GauXC: Williams-Young et al., *Front. Chem.* 8:581058 (2020) & *JCP* 158, 234104 (2023); SSF: Stratmann-Scuseria-Frisch, *CPL* 257, 213 (1996); pruned grids: Dasgupta-Herbert, *JCC* 38, 869 (2017); IncDFT: Q-Chem). Happy to share the full brief with the per-method numbers.</sub>

karmachoi:
Thanks @chatgpt-codex-connector — both findings were real and are fixed in `9f459480f`:

1. **UMRSF FP32 sizing** — `parallel_start` now allocates the FP32 buffers only for the base `int2_mrsf_data_t` (via `select type`) and sizes `d3_sp` by `nmatrix`. UMRSF's own update doesn't use the FP32 accumulator, so FP32 is left inert (exact FP64) there rather than aborting on the non-conformable copy.
2. **Pass-2 (CAM) Fock terms** — `parallel_stop` now reduces the FP64 accumulator across threads *first*, then *adds* the FP32 accumulator, instead of overwriting `f3` with `f3s`. This preserves the `cur_pass==2` short-range-exchange contributions and is still exact for the FP32 pass-1-only case (the FP64 slabs are zero there).

Validation: UMRSF+FP32 now runs and matches FP64 exactly (0 µeV); CAM-MRSF+FP32 matches FP64 to 1 µeV; and a 384-bf MRSF run (c20h42) matches FP64 to 1 µeV at scale.

karmachoi:
Added a third opt-in (commit `0792b634e`): **`OQP_MRSF_RESP_CUTOFF`** — loosen the 2e integral cutoff for the *response* build only.

Rationale: the response is built on converged orbitals, so it tolerates a far looser screening cutoff than the SCF default (`int2e_cutoff = 5e-11`). Loosening it to 1e-7 drops integrals the response cannot resolve, cutting the integral *count* (ERI eval + digestion), and is essentially exact.

Measured (BHHLYP/cc-pVDZ, response int2): cutoff 1e-7 → **~9% (benzene) / ~21% (thymine)**, excitation energies bit-identical (≤1 µeV), Davidson iterations unchanged; 1e-6 → ~16%/~33% at ~3–12 µeV. Effect grows with system size (count reduction, not per-integral cost) and stacks with the FP32 option. Default unset = exact/unchanged; restored after the response so SCF and later steps are untouched.

So the branch now offers three independent, default-off levers over the exact baseline: Coulomb-8→4 digestion (default, exact, ~16%), `OQP_MRSF_FP32` (~+6–8%, ~µeV), and `OQP_MRSF_RESP_CUTOFF` (~9–33% depending on threshold/size, µeV). Happy to split the cutoff one into its own PR if preferred.

karmachoi:
@codex review

Please re-review the latest head (commit `0792b634e`) — in particular the new third opt-in added since your last review: `OQP_MRSF_RESP_CUTOFF`, which loosens `infos%control%int2e_cutoff` for the MRSF response `int2` build only (saved/restored around `int2_driver%init` in `source/modules/tdhf_mrsf_energy.F90`), default unset = exact/unchanged.

karmachoi:
Fixed in `724ed3b55`: widened `rc_env` to `character(len=64)` so long-form `OQP_MRSF_RESP_CUTOFF` values aren't truncated by `get_environment_variable`. Verified short (`1e-7`) and long (`1.000000000000e-7`) forms now give bit-identical results (0 µeV).

karmachoi:
Heads-up — **default change** in `99e092901`: the MRSF *response* integral cutoff now **defaults to 1e-8** (was opt-in via `OQP_MRSF_RESP_CUTOFF`, default = exact SCF 5e-11).

Rationale: the response is built on converged orbitals, and 1e-8 is exact to the printed precision (≪1 µeV — well below both the ~5e-5 regression tolerance and the 1e-6 conv tolerance), so it's a safe default for a modest free speedup that grows with system size. Validated: default(1e-8) matches the old tight result to 0 µeV, iterations unchanged.

Still fully overridable via `OQP_MRSF_RESP_CUTOFF`: looser (1e-7/1e-6) for more speed at µeV cost, or **5e-11 to recover the exact-tight build**. `max(SCF cutoff, requested)` never tightens below the SCF integrals; restored after the response so SCF/gradient steps are untouched.

So the branch's defaults are now: digestion (exact, ~16%) + response cutoff 1e-8 (≪µeV) on by default; FP32 still opt-in.

karmachoi:
Cherry-picked the dftd4→LAPACK static-link-order build fix (`71c5da0` from #232) into this branch to unblock CI — the OFF `gcc-build` failures here were that same main-side linker bug (the one DDPCM example exercises the dispersion path needing `ssytri_`), not the MRSF response code. Will dedup when #232/main lands it.

One refinement vs the original: I guarded the RESCAN group with **`NOT APPLE`**. `$<LINK_GROUP:RESCAN>` is GNU-ld-only — Apple's ld64 rejects it for the CXX link language, and it already resolves the cross-refs by global archive scan, so without the guard macOS + NetLib-static source builds fail to *configure* ("RESCAN ... not supported"). The same `NOT APPLE` guard should be applied on #232. Verified the lib loads and MRSF runs on macOS after the change.

PR #242, opened 2026-06-27
feat: DFT XC-grid compute-once/reuse — collocation-Φ cache (+ experimental IncDFT)

karmachoi:
## What

Two **default-OFF**, env-gated optimizations of the SCF DFT cost. OpenQP already
reuses the J/K (HF) part of the Fock matrix incrementally across SCF iterations,
but the **XC** contribution was rebuilt from the full density every iteration.
These target that gap. Full writeup: `docs/dft_xc_reuse.md`. One commit, off by
default — no behavior change unless an env var is set.

## Opt 1 — collocation-Φ cache (`OQP_XC_PHI_CACHE=1`) — exact

`Φ[μ,i] = φ_μ(r_i)` (+ grid derivatives, weights, per-slice significant-AO pruning
metadata) depends only on geometry + basis + grid + threshold, **not** the density,
yet `run_xc` recomputed it (`compAOs`/`pruneAOs`) on every Fock build. This caches
the post-pruning Φ block per grid slice once per geometry and replays it on later
SCF iterations.

- New module `dft_gridint_phi_cache.F90`; build/replay branch in `run_xc`;
  `resetPrunedPointers` gains a `gather=.false.` mode (replay re-does only the
  density-dependent wavefunction compression, skips the geometry-only AO gather).
  Integrates with the symmetry-reduced weight path.
- Opt-in **only** from the repeated SCF Fock build (`dmatd_blk` and the
  density-driven `dmatd_density_blk` that `calc_fock` uses) — one-shot consumers
  (gradients, response) never cache. Validity keyed by geometry hash + derivative
  order + DFT threshold ⇒ rebuilds transparently on any change.
- Replayed Φ is **bit-for-bit identical**, so the converged energy is unchanged.

**Validation** (converged energy, cache 1 vs 0; 6-31G(d), 4 threads):

| system / functional | ΔE (Ha) | iters | build → replay |
|---|---|---|---|
| benzene / B3LYPV5 | 0.0e+00 | 9 → 9 | 1 build, 8 replays |
| benzene / M06-2X (meta-GGA) | 0.0e+00 | 9 → 9 | 1 build, 8 replays |
| benzene / PBE | 0.0e+00 | 9 → 9 | 1 build, 8 replays |
| C20H42 / PBE | 0.0e+00 | 12 → 12 | 1 build, 11 replays |

**Speedup** (removes ~85–90 % of the `compAOs` collocation cost; total benefit
scales with the collocation share of the XC build):

| case | XC build/iter (off→replay) | XC frac of Fock | total wall |
|---|---|---|---|
| benzene PBE | 0.046 → 0.034 s (−26 %) | ~39 % | 1.73 → 1.62 s |
| C20H42 PBE | 1.07 → 0.93 s (−13 %) | ~34 % | 37.7 → 36.5 s |

Memory trade (why OFF by default): cache = `Σ_slices nAO_pruned·nPts·nAOVecs·8 B`
(C20H42 PBE default grid ≈ 4.8 GB), logged as `cacheMB` under `OQP_XC_TIMING`.
The win grows with coarser grids, meta-GGA, MO-driven builds, and when J/K is
reduced (composes with 2e-integral screening).

## Opt 2 — IncDFT (`OQP_XC_INCDFT=1`) — **experimental, measured non-improvement, OFF**

Whole-matrix reuse of `V_xc[D_ref]`/`E_xc[D_ref]` in a late-SCF DIIS window with
periodic forced full rebuilds (tied to the existing incremental refresh) + exact
final builds. **It does not help** — energy stays within 1e-7 but convergence
blows up:

| system | baseline iters | IncDFT iters | reuses | ΔE |
|---|---|---|---|---|
| benzene / PBE | 9 | 18 | 7 | 0.0e+00 |
| C20H42 / PBE | 12 | 31 | 16 | −7e-10 |

Root cause: XC is nonlinear and the reused matrix is *global*; freezing it yields
a Fock inconsistent with the evolving density, so **each reuse ≈ +1 SCF iteration**,
which always costs more than the skipped XC build (J/K ≥ XC per iter). The
**correct** approach is per-batch ΔP screening / reference subtraction
`V = V_ref + Σ_active (V_s[D] − V_s[D_ref])`, which keeps the Fock accurate ⇒ no
convergence penalty (future work; builds on the Φ cache). Retained here OFF and
clearly experimental as the gating/reference-store scaffolding + reproducible
negative result (`[SCFTIME] ... xc_reused=T/F`).

## Notes for reviewers

- `OQP_XC_TIMING=1` adds per-iteration `[SCFTIME]` (J/K vs XC wall split) and
  `[XCTIME]` (geom-Φ vs density-driven) diagnostics.
- Built/validated with `-DLINALG_LIB_INT64=OFF` (LP64 Accelerate) on macOS.
- **Opt 1 is the part proposed for merge on its merits**; Opt 2 is included only
  as flagged experimental scaffolding (happy to split it out or drop it).


karmachoi:
Addressed both Codex findings (force-pushed; PR is now split into two commits — Opt 1 Φ-cache, Opt 2 experimental IncDFT):

**P2 — Φ-cache opted in too broadly (Hessian path).** Correct: the SCF reaches the grid via `dmatd_density_blk` (density-driven `calc_fock`), while `dmatd_blk` is the one-shot path used by the finite-difference Hessian (`dftexcor`). The opt-in is now set **only** in `dmatd_density_blk`; `dmatd_blk` no longer enables the cache, so the per-displacement Hessian XC builds no longer allocate/retain a never-replayed cache. Verified the SCF still caches (bit-exact, 1 build + N replays) and a numerical-Hessian run completes with the cache on.

**P2 — exact final XC build not guaranteed under IncDFT.** Correct. The SCF convergence test now refuses to accept a converged iteration whose XC was reused (`.and. .not. xc_reuse_now`), forcing a full XC build on the converging iteration (at most one extra iteration; no-op when IncDFT is off). The converged Fock/energy therefore always come from a full build; validated the final iteration reports `xc_reused=F`.

Thanks for the review.

PR #245, opened 2026-06-28
perf(grad): default OQP_GRAD_CUTOFF to 1.0d-8 (follow-up to #244)

karmachoi:
## Summary

Follow-up to #244 (the opt-in `OQP_GRAD_CUTOFF` lever). #244 kept the historic, unusually tight `1.0d-10` Schwarz block cutoff as the default; this **flips the default to `1.0d-8`**.

The gradient is built at the **converged** density, so the derivative-2e screening — the dominant gradient cost (92–99.7% for HF/DFT/MRSF, all via the same `grd2_driver_gen`) — tolerates a looser default.

## Why `1.0d-8`

- **Size-robust accuracy:** max\|ΔG\| ≤ ~1e-6 a.u. vs the tight `1.0d-10` baseline through the 36-atom systems tested (benzene→pentacene, HF/DFT/MRSF) — well under the 1e-5 budget and near the 1e-6 ideal.
- **Speed:** ~5–12% off the dominant 2e-derivative build, growing with system size.
- **References unaffected:** the drift on the small-molecule example references is ~1e-7, ~500× below the regression-suite tolerance (≈5e-5). The HF/DFT/MRSF/TDDFT gradient and optimize example regressions all pass unchanged at the new default.

## Behavior / overrides

- Default `OQP_GRAD_CUTOFF = 1.0d-8`.
- `=1.0d-10` restores the historic tight screening **byte-for-byte**.
- `=1.0d-7` is more aggressive (~10–25%) but max\|ΔG\| exceeds 1e-5 past ~18-atom HF (DFT/MRSF tolerate it further).
- `OQP_GRAD_STATS=1` prints screening counts (off by default).
- `grd2_hess_driver` (analytic Hessian) is unchanged at `1.0d-10`.

## Validation

Gradient error vs the `1.0d-10` baseline within budget per method/size; analytic gradient cross-checked vs finite difference (2.4e-5, FD-truncation-limited); geometry optimization converges in identical step counts (final energy Δ 6e-10). See `GRAD_SCREENING_NOTES.md` for the full table.

### DIFF
--- a/GRAD_SCREENING_NOTES.md
+++ b/GRAD_SCREENING_NOTES.md
@@ -7,8 +7,12 @@ MRSF-TDDFT excited state) by tolerating looser settings on work that is
 recomputed at the **converged** density, validated against gradient accuracy
 (Hartree/Bohr), which is stricter than energy.
 
-All changes are **default-OFF** (env-gated). With no env set the code path is
-byte-identical to baseline.
+**Default `OQP_GRAD_CUTOFF = 1.0d-8`** (the historic Schwarz cutoff was an
+unusually tight `1.0d-10`). The env var overrides it; **`=1.0d-10` restores the
+historic tight screening byte-for-byte**. `1e-8` is the size-robust choice
+(max|ΔG| ≤ ~1e-6 a.u. through the 36-atom systems tested) and sits well below
+the regression-suite tolerance (≈5e-5, max-element deviation rounded to 4
+decimals), so it does not perturb the small-molecule example references.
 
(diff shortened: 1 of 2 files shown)


karmachoi:
Closing in favor of the shipped opt-in lever (#244). The capability is available via `OQP_GRAD_CUTOFF=1e-8` (or `1e-7`); flipping the *default* from 1.0d-10 to 1.0d-8 is a project-policy call (it changes default gradient numerics by ~1e-6 for ~5-12%) better left to maintainers. The validation supporting that one-line change is documented in GRAD_SCREENING_NOTES.md (merged via #244).

PR #247, opened 2026-06-28
feat(perf): user-facing `perf` preset + input keys for performance options (+ gfortran-12 tagarray build fix)

karmachoi:
## Summary

Adds a user-facing performance preset **`[input] perf = 0|1|2|3` (default `1`)** that bundles
OpenQP's performance knobs into one accuracy↔speed dial, plus the per-knob **input keys** behind
it — all routed through the shared `control` struct, **with no environment variables**. Also fixes
a build break on `main` under gfortran ≥ 12 (separate first commit, cherry-pickable).

Outcome of an audit + CPU benchmark of the recent perf machinery (C2F XC grid, Φ-cache, progressive
screening, gradient/response cutoffs, MRSF FP32, z-vector warm-start).

## Commits
1. **`fix(build)`: tagarray v1.0.0 under gfortran ≥ 12.** A 144-char `iso_c_binding` USE line in
   `container.f90` is truncated by tagarray's own `-std=f2018` → `pip install .` of `main` fails on
   GCC ≥ 12. Fixed with an idempotent `PATCH_COMMAND` (mirror upstream in `Open-Quantum-Platform/tagarray`).
2. **`feat(perf)`: the `perf` preset + input keys (control-struct, no env vars).**

## Design
Each knob is a `control_parameters` field (`source/types.F90` + the C mirror in `include/oqp.h`),
set from Python via the existing setter dispatch, and read directly from `infos%control` in Fortran —
the previous `get_environment_variable("OQP_*")` reads for these knobs are removed.

New input keys (sentinel default `auto` = defer to the preset; any explicit value overrides it):

| section | keys |
|---|---|
| `[scf]` | `xc_c2f`, `xc_phi_cache`, `xc_incdft`, `grad_cutoff` |
| `[tdhf]` | `resp_cutoff`, `fp32`, `zv_warmstart` |

Resolved in `mol.load_config`, filling only keys still at `auto`. Precedence:
**control default < `perf` preset < explicit input key**. Set `perf=-1` to disable the preset.

## The levels (CPU-calibrated: MKL + Apple Accelerate)

| `perf` | meaning | speed | accuracy |
|---|---|---|---|
| 0 | strict reference / reproducible | baseline | bit-reference (fixed thread count) |
| **1 (default)** | recommended production — exact wins only (resp 1e-8, z-vector warm-start, + Fock digest) | ~6% faster | energies/VEEs **exact** |
| 2 | faster, tiny degradation — + C2F grid + grad cutoff 1e-8 | — | gradients ≤ 5e-7 a.u. |
| 3 | aggressive, degradation allowed — grad 1e-7, response 1e-6 | **~17% faster** (MRSF energy) | ~6 µeV on VEEs |

Φ-cache, IncDFT, FP32 and progressive screening were **net-negative/experimental on CPU** (FP32 ~2×
slower under MKL) and are enabled by **no preset** — kept as explicit input keys for other regimes (GPU XC).

## Benchmark highlights (chc4 Xeon 8368 / MKL; cross-checked on M3 Ultra / Accelerate)
- All settings reproduce reference energies/VEEs to printed precision (C20H42 = -786.7633652204).
- "Enable everything" is net-slower for single points (C2F × pscreen adds 3–10 SCF iterations);
  Φ-cache −8%, FP32 ~2× slower — hence excluded from presets.
- `perf=3` thymine MRSF energy: **37.5 s → 31.1 s (1.21×)** for ~6 µeV.

## Review feedback addressed
- **`auto` is no longer treated as an override** — the preset is applied even when a fully
  default-expanded config dict is passed (value-based resolution; regression test added).
- **z-vector warm-start is refreshed every call**, so a per-calculation override applies in
  long-lived (Python) workflows.

## Tests / verification
- `tests/test_perf_levels.py` — 7 cases (level mapping incl. expanded-dict, override precedence,
  validation). ALL PASS.
- Clean-checkout build on chc4 (no-libint, MKL ILP64, OpenMP): builds, imports, `perf=0..3` exact/
  controlled, explicit keys override the preset, **no `OQP_*` env var governs the path**, `perf=1`
  is the default. CI: gcc-build, sdist, and Linux wheel-smoke all green.

### DIFF
--- a/docs/perf_levels.md
+++ b/docs/perf_levels.md
@@ -1,0 +1,82 @@
+# Performance options and the `perf` preset
+
+OpenQP's performance knobs are ordinary **input keys** backed by the shared control
+struct — there are no environment variables involved. A single opt-in preset
+`[input] perf` bundles them into one accuracy↔speed dial.
+
+```
+[input]
+perf = 1        # 0, 1, 2, or 3 (default 1; set perf=-1 to disable the preset)
+```
+
+## The four levels
+
+| `perf` | Use it for | What it does | Accuracy |
+|---|---|---|---|
+| **0** | strict reference / reproducibility | every accelerator off, every cutoff tightest | bit-reference (fixed thread count) |
+| **1** | **recommended production** | only exact, proven-helpful knobs: MRSF response cutoff `1e-8`, z-vector warm-start (+ always-on Fock digestion) | ≈ reference (≤ µEh) |
+| **2** | faster, tiny degradation | `perf=1` + coarse-to-fine XC grid + gradient Schwarz cutoff `1e-8` | gradients within ~5×10⁻⁷ a.u.; SCF exact at convergence |
+| **3** | aggressive, **degradation allowed** | looser cutoffs traded for speed: gradient `1e-7` (~10⁻⁵ a.u.), response `1e-6` (~few µeV) | small, controlled — verify for your system |
+
+**`perf=1` is the default** (recommended production, exact). Set `perf=-1` to disable the preset
+and leave every knob at its control default. All levels were calibrated
+against a CPU benchmark (MKL + Apple Accelerate) across HF/DFT/TDDFT/MRSF energies and MRSF
+gradients.
+
+> **Why these and not the rest.** On the CPU builds benchmarked, the reliable speedups are
+> the `perf=1` knobs (response cutoff, warm-start, Fock digestion) and, at higher levels,
+> *looser integral cutoffs*. The XC Φ-cache, IncDFT, MRSF FP32 and progressive screening
+> (`pscreen`) were performance-neutral to *negative* on CPU (FP32 ~2× slower under MKL), so
+> **no preset enables them** — they remain available as explicit input keys for regimes the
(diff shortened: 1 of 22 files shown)


karmachoi:
Thanks @chatgpt-codex-connector — both points fixed in 365391f6:

1. **`auto` no longer counts as an override.** Resolution is now value-based: a perf key is
   filled from the preset whenever its value is the sentinel `auto`, so a fully default-expanded
   config (e.g. from the scripting API / `OQPConfigParser.validate()`) still receives the preset.
   Only a non-`auto` value blocks it. Added a regression test that resolves from an all-`auto`
   expanded dict. (`pscreen` — the one non-`auto` key — was also net-negative on CPU, so it is
   no longer in any preset; it stays an explicit opt-in.)
2. **z-vector warm-start is refreshed every call.** `zv_warm_on = (infos%control%mrsf_zv_warmstart /= 0)`
   now runs before the `zv_cfg_init` guard, so a later job's `[tdhf] zv_warmstart` / `perf` override
   applies in long-lived workflows; the remaining env-only levers stay one-shot.

PR #255, opened 2026-07-03
ci(policy): enforce the four PR contribution rules

karmachoi:
## Summary

Introduce automatic enforcement of four PR contribution rules, using a mix of a deterministic CI gate and the existing Claude/Codex automated reviews.

The four rules (documented in the new **`AGENTS.md`**):
1. **BLAS/LAPACK** must go through the `oqp_linalg` wrapper — no raw `dgemm`/`dgels`/`dgetrf`/... outside `source/mathlib/lapack_wrap.F90` / `blas_wrap.F90`; new routines are registered in `OQP_ACCELERATE_ILP64_SYMS`.
2. **New functionality** ships with a small, fast example under `examples/`.
3. **User-facing additions** get a pythonic Python-API addition in `pyoqp/oqp/openqp.py` + a `tests/test_openqp_api.py` test.
4. **New keywords/workflows** are documented in openqp-docs (a linked docs PR).

## How each rule is enforced

| Rule | Mechanism |
| --- | --- |
| 1 | **Hard gate** — `tools/check_blas_wrapper.py` run by the new `PR policy` workflow (`.github/workflows/pr-policy.yml`), plus the existing macOS `check_accelerate_aliases` link-time check |
| 2 | `openqp --check_feature_coverage` (existing) for opt-in flags + reviewer check for the rest |
| 3 | Claude/Codex review against `AGENTS.md` + `tests/test_openqp_api.py` |
| 4 | PR-template checklist (cross-repo) + reviewer check |

- **`AGENTS.md`** is read natively by the Codex reviewer; the **Claude review** prompt (`.github/workflows/claude.yml`) now points at it and reports PASS / what's missing per rule.
- **`.github/pull_request_template.md`** surfaces the rules as an author checklist — the practical lever for the cross-repo openqp-docs rule.

## The Rule-1 lint

`tools/check_blas_wrapper.py` is a portable, dependency-free source lint: it fails when any BLAS/LAPACK routine referenced from `source/` (outside the wrapper modules) is missing from `OQP_ACCELERATE_ILP64_SYMS` — meaning it has no `oqp_*_i64` wrapper and no macOS Accelerate ILP64 alias. It is the every-platform, pre-build complement to the macOS-only, link-time `cmake/check_accelerate_aliases.cmake` gate.

Verified:
- **Green on current `main`** — "every BLAS/LAPACK routine referenced in source/ is registered (165 symbols)", zero false positives across the ~90 bare-`dgemm` call sites.
- **Catches the real bug** — removing `dgels`/`dgetrf` from the list reproduces exactly `source/modules/qmmm.F90:1357/1362: dgels` (the un-wrapped call that shipped in #205 and failed only the macOS Accelerate build).

## Notes / follow-ups

- The AI reviews are **non-blocking** (they comment); the deterministic gates block. That's why Rule 1 (mechanically checkable) is a hard gate while Rules 3–4 (subjective / cross-repo) are reviewer + checklist.
- Extending `--check_feature_coverage` to non-boolean additions (new `runtype`/`[section]`/`method`) would strengthen Rule 2 — left as a follow-up.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,12 +7,13 @@
 See [AGENTS.md](../AGENTS.md) for details. The automated Claude and Codex
 reviewers check these too.
 
-- [ ] **BLAS/LAPACK** — no raw `dgemm`/`dgels`/`dgetrf`/... calls; all go through
-      the `oqp_linalg` wrapper (raw calls only inside
-      `source/mathlib/lapack_wrap.F90` / `blas_wrap.F90`).
+- [ ] **BLAS/LAPACK** — every referenced BLAS/LAPACK routine is registered in
+      `OQP_ACCELERATE_ILP64_SYMS` (`cmake/oqp_functions.cmake`); new code prefers
+      `use oqp_linalg`. Unregistered routines fail the `PR policy` lint.
 - [ ] **Example** — new functionality (runtype/method/`[section]`/keyword/flag)
-      adds a small, fast example under `examples/` (passes
-      `openqp --check_feature_coverage`).
+      adds a small, fast example under `examples/`. (New opt-in **boolean** flags
+      are gated by `openqp --check_feature_coverage`; other additions are checked
+      by the automated reviewers.)
 - [ ] **Python API** — user-facing additions are usable from `oqp.openqp.OpenQP`
       and covered by a `tests/test_openqp_api.py` test.
 - [ ] **Docs** — user-facing keywords/workflows documented in openqp-docs; link
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,21 @@
+<!-- Describe the change and link any related issues. -->
+
+## Summary
+
+## Contribution checklist
+
+See [AGENTS.md](../AGENTS.md) for details. The automated Claude and Codex
+reviewers check these too.
+
+- [ ] **BLAS/LAPACK** — no raw `dgemm`/`dgels`/`dgetrf`/... calls; all go through
+      the `oqp_linalg` wrapper (raw calls only inside
+      `source/mathlib/lapack_wrap.F90` / `blas_wrap.F90`).
+- [ ] **Example** — new functionality (runtype/method/`[section]`/keyword/flag)
+      adds a small, fast example under `examples/` (passes
+      `openqp --check_feature_coverage`).
+- [ ] **Python API** — user-facing additions are usable from `oqp.openqp.OpenQP`
+      and covered by a `tests/test_openqp_api.py` test.
+- [ ] **Docs** — user-facing keywords/workflows documented in openqp-docs; link
+      the companion PR here: <!-- openqp-docs PR # -->
+
+<!-- If a checklist item does not apply, say why. -->
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -53,10 +53,12 @@ jobs:
             In addition to correctness/bug review, check the PR against the four
             contribution rules in AGENTS.md (in the repo root) and report each
             explicitly:
-              1. BLAS/LAPACK calls go through the oqp_linalg wrapper (no raw
-                 dgemm/dgels/dgetrf/... outside source/mathlib/lapack_wrap.F90
-                 and blas_wrap.F90); new routines are registered in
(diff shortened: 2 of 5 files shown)


karmachoi:
Thanks @chatgpt-codex-connector — addressed in `2d7dadd1`.

**P3 (scan the linked test harnesses) — fixed.** The gate now scans the
`tests/fortran/*.F90` bind(C) self-test harnesses that `source/CMakeLists.txt`
appends to the `oqp` library target, in addition to `source/`. A raw BLAS/LAPACK
call planted in `tests/fortran/hess1_selftest.F90` is now flagged.

**P2 (detect LAPACK calls outside the hard-coded list) — fixed.** Added a
list-free heuristic alongside `CANON`: a `call NAME(` whose NAME has the LAPACK
`[precision][matrix-type][op]` shape, is **not** defined anywhere in the scanned
source (i.e. it resolves to an external BLAS/LAPACK symbol), and is not
registered, is now flagged even when CANON doesn't list it. It's anchored on the
real LAPACK matrix-type digraphs (`ge`/`sy`/`he`/`po`/`tr`/…) rather than "[sdcz]
+ letters", which is what keeps ordinary externals like `call solver(` (the
OpenTrustRegion entry point) and `call second(` from being mistaken for LAPACK.
Verified it now catches a new `call dgesdd(`/`call dpotrs(`/`call dsyevr(`. (BLAS
L1/L2/L3, which don't follow that scheme, remain fully covered by CANON.)

**P1 (reject registered raw BLAS calls) — intentional, now documented.** This is
a deliberate design choice rather than a gap. OpenQP is **ILP64-only**
(`-fdefault-integer-8`, `BLA_SIZEOF_INTEGER=8`, CMakeLists.txt:228-234), so a
bare `call dgemm(...)` binds correct-width 8-byte integers on Linux; the *only*
failure mode is the macOS Accelerate alias, which registration in
`OQP_ACCELERATE_ILP64_SYMS` fixes. Several long-standing sites (e.g.
`scf_addons.F90`) call bare BLAS directly and are correct on this build — so the
enforced minimum is **registration**, not literal `use oqp_linalg`. Enforcing
wrapper-usage would false-positive on ~90 correct legacy call sites. New code
should still prefer `use oqp_linalg` for clarity; `AGENTS.md` and the
Claude-review prompt now state this distinction explicitly.

The macOS `check_accelerate_aliases` gate (which nm's the real linked library)
remains the exhaustive backstop; this portable lint is the fast pre-build
early-warning that also runs on Linux-only PRs.

karmachoi:
Ran an additional internal adversarial review of all 5 files in this PR and fixed
four confirmed issues in `9c2a8f7e` (5 other candidate findings were checked and
refuted as non-defects):

1. **Lint false-negative (`search` → `finditer`).** `main()` used
   `call_rx.search`, which only inspects the *first* `call` on a line. Fortran
   `;` statement separators allow multiple calls per line (a style already in the
   tree, e.g. `hf_hessian.F90`), so a non-first unregistered BLAS call could slip
   through. Now iterates every `call` on the line. Verified it catches
   `call setup(x); call dpotrs(...)`.

2. **CI-integrity: gate ran the PR's own copy of the script.** `pr-policy.yml`
   executed the PR-head `tools/check_blas_wrapper.py`, so a PR could weaken rule 1
   by editing the gate script in the same PR. The script now takes an optional
   `REPO_ROOT` arg, and the workflow runs the **trusted base-branch copy** of the
   script against the PR-head tree (bootstrap fallback to the in-tree copy only
   when the base has no gate yet). Job is `contents: read`, no secrets, and only
   statically reads code as text. The base-branch link-time
   `check_accelerate_aliases.cmake` gate remains the independent backstop.

3. **PR-template rule-1 wording too strict.** The checkbox said "raw calls only
   inside the wrapper modules", contradicting AGENTS.md and the actual gate (which
   enforce *registration*, not literal `use oqp_linalg`; ~90 legacy bare-BLAS
   sites are correct by design). Reworded to the registration invariant.

4. **PR-template rule-2 over-claimed `--check_feature_coverage`.** It implied that
   gate covers runtypes/methods/sections/keywords, but it only gates opt-in
   **boolean** flags. Scoped the parenthetical honestly and pointed the rest at
   the automated reviewers. Aligned the matching AGENTS.md wording too.

PR #258, opened 2026-07-04
feat(qmmm): optional RCD/RC/Z1 frontier-charge redistribution for covalent QM/MM boundaries

karmachoi:

PR #259, opened 2026-07-04
fix(qmmm): sort qm_atoms into topology order so QM forces scatter correctly

karmachoi:
## Bug

The QM/MM driver builds the QM geometry by iterating `topology.atoms()` filtered
by membership — i.e. in **topology order** (ascending `atom.index`). So the QM
gradient, coupling force, and ESP charges returned by the engine are in topology
order.

But the force-scatter loops (`_assemble_force`, `_assemble_force_espf`) and the
link-atom `host_row` projection index those arrays **by position in
`self.qm_atoms`**. If `qm_atoms` is supplied out of ascending order (e.g.
`qm_atoms = "5,2,7"`), those positions don't match topology order, so the QM
gradient / coupling force / link-atom projection land on the **wrong atoms** —
a silent wrong-number (no crash).

This is pre-existing (independent of, and predating, the frontier-charge work in
#258, which inherits the same scatter).

## Fix

Sort `qm_atoms` into ascending (topology) order at construction. This does **not**
change the QM calculation — the engine already sees the atoms in topology order —
it only realigns the force scatter. A shuffled selection now normalizes to
ascending order.

Regression test in `tests/test_qmmm_qm_atoms_order.py` (OpenMM/backend-gated).

### DIFF
--- a/pyoqp/oqp/library/qmmm_driver.py
+++ b/pyoqp/oqp/library/qmmm_driver.py
@@ -459,7 +459,11 @@ class OpenQpQMMM:
         self.positions = positions
         self.topology = topology
         self.mm_systems = mm_systems
-        self.qm_atoms = qm_atoms
+        # Re-normalize to ascending (topology) order on every call: callers
+        # (e.g. QMMM_MD) pass their own config-order qm_atoms here, which would
+        # otherwise overwrite the sorted copy set in __init__ and reintroduce the
+        # force mis-scatter this fix prevents. See __init__ for the rationale.
+        self.qm_atoms = np.array(sorted(int(i) for i in qm_atoms), dtype=int)
 
         potmm = potqm = None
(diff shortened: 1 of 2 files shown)


karmachoi:
Good catch — fixed in the latest commit. `compute_force` reassigns `self.qm_atoms` from its argument (and `QMMM_MD` passes its own config-order copy every step), which overwrote the `__init__` sort and reintroduced the mis-scatter. `compute_force` now re-normalizes to ascending (topology) order as well, so the driver always assembles forces in topology order regardless of caller.

I deliberately did **not** sort `QMMM_MD.self.qm_atoms` itself: it is used by `_apply_xyz_positions` to map an optional `qm_atoms_xyz` / `qm_list` correspondence (order-sensitive), and the driver builds its mol geometry in topology order independently — so normalizing inside the driver (both `__init__` and `compute_force`) is the safe, sufficient fix.

PR #290, opened 2026-07-20
Add physical-root MRSF state analysis

karmachoi:
Summary:
- correct fragment CT analysis to use the physical MRSF root-pair 1-TDM
- add NTO-based LE/CT and n/pi/sigma to pi*/sigma* fractional analysis
- expose a public Python API, tests, and a small example

Companion PRs:
- openqp-dftb native API: https://github.com/Open-Quantum-Platform/openqp-dftb/pull/25
- Manual: https://github.com/Open-Quantum-Platform/openqp-docs/pull/20

Validation:
- python3 -m pytest -q tests/test_excited_toolkit.py tests/test_mrsf_state_analysis.py (11 passed, 2 skipped)

### DIFF
--- a/examples/MRSF-TDDFT/STATE_ANALYSIS.py
+++ b/examples/MRSF-TDDFT/STATE_ANALYSIS.py
@@ -1,0 +1,28 @@
+"""Physical-root MRSF state analysis for a finished OpenQP calculation.
+
+Run the MRSF energy job in the current process, then pass its ``mol`` object to
+``report_states``.  Atom and state indices below are zero-based.
+"""
+from oqp.interop import AOBasis, MRSFExcitedStates, analyze_mrsf_transition
+
+
+def report_states(mol, fragments, plane_normal=None):
+    states = MRSFExcitedStates(mol)
+    ao = AOBasis(mol)
+    reports = []
+    for final_state in range(1, states.nstates):
+        report = analyze_mrsf_transition(
+            states, ao, final_state, fragments, ref=0,
+            plane_normal=plane_normal)
+        character = report["orbital_character"]
+        print(
+            f"S0 -> S{final_state}: {character['label']}; "
+            f"LE={report['le_fraction']:.3f}, CT={report['ct_fraction']:.3f}; "
+            f"fractions={character['fractions']}")
+        reports.append(report)
+    return reports
+
+
+# Example after running formaldehyde: one chromophore fragment means LE=1 by
+# construction.  Use chemically meaningful donor/acceptor fragments for CT.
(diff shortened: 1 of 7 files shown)


mohsenkor (approved):
Approving after a two-reviewer pass (mine plus Codex, independently).

The core change is right: dropping `amplitude_matrix` in favour of `tdm_ao(ref, n)` means the analysis describes a physical MRSF root pair rather than a response root relative to the auxiliary high-spin determinant.

Three review findings were fixed on the branch before this approval:

- **pi character was not rotationally invariant.** The projection summed `n_a^2 q_a^2` over Cartesian p components and dropped the cross terms, so a p orbital along a `[1,1,1]` plane normal read as 1/3 pi instead of 1 — a rigid rotation of the same molecule could turn pipi* into mixed or sigmasigma*. It stayed invisible because the ch2o fixture is planar in yz, putting its normal exactly on x. Now projected coherently per shell as `(n.q)^2`. Codex flagged this independently as its top finding.
- **Vanishing root-pair 1-TDMs returned noise as data**: `LE=0.0, CT=0.0, classified=True, label="mixed"` for a symmetry-forbidden pair. Now NaN with `well_defined=False` and an explicit reason.
- **The example never ran** — it only defined a function and referenced an undefined `runner`.

Codex then re-reviewed those fixes by building OpenQP and rotating formaldehyde, which caught a vacuous transpose assertion (single fragment, so Omega was 1x1) and an overclaim: a 5.4e-3 residual anisotropy survives on 6-31G* because AOBasis is Cartesian-only and normalized Cartesian d shells do not rotate orthogonally. That is documented with the measured number rather than claimed away.

Known limitations, accepted deliberately and documented in the code:
- the 5.4e-3 residue above — two orders below the label_threshold margin, but don't read the third decimal of a fraction as physical when polarization functions are present
- `n` is heteroatom in-plane population, not a localized lone pair, so the n/sigma split is indicative while pi is well defined
- `fragment_ct_matrix` breaks compatibility: `amplitude_norm` is gone and `n == ref` now raises. No in-repo callers; openqp-docs#20 should reflect it.

All 17 checks green.

mohsenkor:
Fixed in 
`_fragment_map` now normalizes each fragment to a 1-D int array and checks its length, so `np.flatnonzero`/`np.array_split` fragments work. Reproduced the `ValueError: truth value of an array with more than one element is ambiguous` first and added a regression test covering both the single-atom and multi-atom array cases. Good catch — `analyze_mrsf_transition` builds `hetero_atoms` with `np.flatnonzero` itself, so this was a real trap.

PR #292, opened 2026-07-21
Adopt research and commercial dual licensing for OpenQP

karmachoi:
## Summary

- replace the GPL option for OpenQP versions distributed with the new license with the **OpenQP Research License 1.0**
- establish two alternatives: free qualifying academic/nonprofit research use or a separate written commercial license from Open Quantum Inc.
- require a commercial license for every use by, for, or on behalf of a for-profit entity, including internal R&D, evaluation, benchmarking, proof-of-concept work, sponsored/contract research, products, services, and production
- keep commercial scope, fees, and pricing in private written agreements or quotes
- preserve rights already granted for copies of earlier OpenQP versions validly received under GPLv3, without extending those rights to later versions
- limit the new license to material Open Quantum Inc. owns or has authority to license, preserve third-party terms, and require separate contributor authorization before external code is incorporated
- update the README and Python package metadata so they do not continue to identify current OpenQP as GPLv3

## Why

The previous repository notice combined GPLv3 with academic/noncommercial language. Those terms conflict because GPLv3 permits commercial use and private internal company use. This PR implements the intended prospective model with a source-available research license instead of GPLv3.

The current research option is not an open-source license. Public access to the source does not authorize company use. Commercial pricing is not published in this repository.

## Historical boundary

This change does not retroactively revoke GPLv3 rights for copies of earlier OpenQP versions already received under GPLv3. It applies prospectively to OpenQP copies distributed with the new license. No historical GPL right creates an entitlement to updates, upgrades, later releases, support, trademarks, or separately licensed materials.

## Validation

- `git diff --check`: passed
- `pyproject.toml` TOML parse: passed
- GPL references remaining in the changed files are limited to the historical-license boundary

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,43 +1,674 @@
-# Licensing Terms for Open Quantum Platform (OpenQP)
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
-## Overview
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
 
-The Open Quantum Platform (OpenQP) is distributed under an open-source license for academic and non-commercial use. To sustain its development, a dual licensing model has been adopted, allowing users to choose between a free open-source license or a commercial license for proprietary use. Revenue from commercial licenses directly supports the ongoing development of OpenQP.
+                            Preamble
 
-OpenQP is offered under two distinct licensing options:
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
 
-1. **Open Source License**: GNU General Public License v3 (GPLv3)
-2. **Commercial License**: For proprietary use of OpenQP
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+the GNU General Public License is intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.  We, the Free Software Foundation, use the
+GNU General Public License for most of our software; it applies also to
+any other work released this way by its authors.  You can apply it to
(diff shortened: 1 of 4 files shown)


mohsenkor (commented):
### General note (non-blocking) — governance & ownership as we scale

This license clarification is good on its own. Separately, I want to put a longer-term point on record now, before the codebase and contributor list grow further.

**Copyright ownership / CLA.** Under GPLv3 every contributor keeps copyright in their own contribution — they've only licensed it in under GPLv3. That's fine for the open-source release, but it means no single party currently holds rights to 100% of the code. If OpenQP ever offers **separate commercial / proprietary (dual) licensing** — which `LICENSING.md` already gestures at — we can only license material we have authority over. To make that real we'd want to:

1. **Settle the owning entity** — decide whether the copyright sits with Prof. Choi, KNU, or Open Quantum Inc., and confirm KNU's IP position on faculty-authored code.
2. **Adopt a CLA** — a license-grant CLA (Apache / Project-Harmony style: contributors keep their name but grant relicensing rights). A CLA bot (e.g. CLA Assistant) can gate new PRs automatically; the ~12 existing contributors would do a one-time backfill sign-off. Much easier to do now than after another year of commits.

**On CODEOWNERS (as in cp2k).** Worth adopting too — but it solves a *different* problem. `CODEOWNERS` auto-assigns reviewers/maintainers per file path and encodes who is responsible for which subsystem; it's about review quality and maintainership, **not** copyright. cp2k pairs it with clear contribution rules. We could do the same split: **CODEOWNERS for review ownership + a CLA for legal ownership.** They're complementary, not substitutes.

Also minor, in the same spirit: per-file `SPDX-License-Identifier: GPL-3.0-only` headers and a named copyright line would help provenance and tooling as we grow.

None of this blocks this PR — just flagging it so we plan for it deliberately rather than discover it later.

PR #293, opened 2026-07-21
ci: run the pytest suite

mohsenkor:
## Problem

CI builds OpenQP and runs the example/reference regression suite (`openqp --run_tests all`), which validates numerical references end to end — but it never imports the pure-Python layer.

The `tests/` directory holds **115 test files** covering the analysis toolkit, the interop/export APIs, input handling, the periodic table and more. **None of them were executed by any workflow.** A regression anywhere under `pyoqp/oqp` could land fully green.

This surfaced while reviewing #290: that PR adds two pytest files, and neither would ever have run in CI.

## Change

Run `pytest tests` on the `ubuntu-24.04` non-MPI leg of `gcc-build`, which already has OpenQP installed. One leg is enough — the suite is Python-only, so the arm64 leg and the MPI leg would re-run identical assertions against identical code.

Tests that need a compiled build gate on `OPENQP_ROOT`. The installed package self-locates via `resolve_oqp_root()` and doesn't need the variable, so the step sets it explicitly to the installed package directory — otherwise the one job that *can* run those integration tests would skip them silently.

## Risk

This is the first time these 115 files run in CI, so pre-existing failures may surface. That is the point of the PR; if any appear, they get fixed or explicitly marked before merge. Watch this PR's own CI run for the answer.

https://claude.ai/code/session_011cCWPeiTGrfWYLQKgSy3U9

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -181,7 +181,12 @@ jobs:
          - name: Run Python test suite
            if: ${{ matrix.MPI == 'OFF' && matrix.os == 'ubuntu-24.04' }}
            run: |
-            pip install pytest
+            # xdist: the suite runs real calculations and took 11.4 min serially,
+            # which nearly doubled this job. The example suite above already
+            # saturates the runner cores, so do the same here. --dist loadfile
+            # keeps every test in a file on one worker: several tests build
+            # projects from a shared cwd and would collide if spread apart.
+            pip install pytest pytest-xdist
             # Tests that need a build gate on OPENQP_ROOT. The installed package
             # self-locates via resolve_oqp_root(), so point the variable at the
             # package directory to opt those integration tests in rather than
@@ -194,7 +199,7 @@ jobs:
             # `pyoqp.oqp.periodic_table` through the source tree rather than the
             # installed `oqp` package. Bare pytest only adds tests/ and the
             # module fails to import.
-            python -m pytest tests -q --durations=10
+            python -m pytest tests -q -n auto --dist loadfile --durations=15
 
    # macOS legs: build with PURE DEFAULTS (no LINALG flags) so CI exercises the
    # native-BLAS platform policy exactly as users hit it: LINALG_LIB=auto must
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -172,6 +172,25 @@ jobs:
               exit 1
             fi
 
+         # The example suite above validates numerical references end to end, but
+         # it never touches the pure-Python layer -- the tests/ pytest suite was
+         # not executed by CI at all, so everything under pyoqp/oqp (the analysis
+         # toolkit, the interop/export APIs, the input handling) could regress
+         # silently. Run it on one leg: it is Python-only, so a second OS or the
+         # MPI leg would re-run identical assertions against the same code.
+         - name: Run Python test suite
+           if: ${{ matrix.MPI == 'OFF' && matrix.os == 'ubuntu-24.04' }}
+           run: |
+            pip install pytest
+            # Tests that need a build gate on OPENQP_ROOT. The installed package
+            # self-locates via resolve_oqp_root(), so point the variable at the
+            # package directory to opt those integration tests in rather than
+            # letting them skip silently in the one job that can run them.
+            OPENQP_ROOT="$(python -c 'import oqp, os; print(os.path.dirname(oqp.__file__))')"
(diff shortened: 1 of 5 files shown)


mohsenkor:
## First real run: the suite executes, and ~22 tests fail

The `sys.path` fix (`python -m pytest`) got collection working. ~1000 tests now run; **~22 fail**. All of these are **pre-existing** — none are caused by this PR, they were simply never executed. Triage by root cause:

**1. `TypeError: uint32() takes no arguments` — 13 failures**
Raised from `np.random.RandomState(seed)` inside `numpy/random/bit_generator.pyx`. Hits `test_symmetry_mo_label_wiring.py`, `test_symmetry_mode_labels.py` and others. This is an environment/ABI problem, not test logic — it needs a numpy-version diagnosis on the CI image.

**2. Genuine API drift — ~8 failures**
- `ValueError: cannot create an OBJECT array from memory buffer` (×4)
- `AttributeError: 'object' object has no attribute 'buffer'` (×2)
- `AttributeError: Key 'OQP::VEC_MO_B' not found in QOP data` (×2)
- `AttributeError: 'MPIManager' object has no attribute 'use_mpi'` (×2)

These look like real breakage between the tests and the code they exercise.

**3. Stale source-content guard tests — ~5 failures**
Tests asserting on literal file contents that have since drifted:
- README contributor list no longer contains `VladimirMakhnev`
- `external/CMakeLists.txt` no longer contains `string(MD5 _oqp_external_make_program_hash ...)`
- the `libtagarray-v0.0.6-default-integer-shapes.patch` reference is gone
- `ij_pair = 0` no longer in the int2 Fortran source
- `'from pyscf'` now unexpectedly present in the FCIDUMP export module

**4. Missing optional dependency — 1 failure**
`ModuleNotFoundError: No module named 'qcelemental'` (should be an `importorskip`).

Also of note: the analytic-vs-numerical Hessian tests (`RohfAnalyticHessian`, `RoksAnalyticHessian`, `RhfDFunctionAnalyticHessian`) are among the failures and deserve a look on their own merits.

**This PR is not mergeable as-is** — it would turn CI red. Holding for a decision on how to sequence the cleanup.

mohsenkor:
## Correction to the triage above, and final state: green

**The "38 failures" I reported earlier was wrong.** Most of it was an artifact, and the cause is worth knowing about independently of this PR.

### The suite is not safe in a shared interpreter

39 test files mutate `sys.modules`, and `tests/test_analytic_hessian.py:58` replaces the entire `oqp` package with a stub:

```python
sys.modules["oqp"] = oqp_stub
```

Whatever runs after that in the same process then fails with `ImportError: cannot import name 'perf_levels' from 'oqp.utils' (unknown location)` and similar. So **the set of failures depended on execution order** — the serial run and the parallel run disagreed on 22 tests, and neither was telling the truth.

It also explains the 683s → 27s "speedup" the xdist commit appeared to buy: those tests weren't running fast, they were dying at import before doing any work.

Fixed with `--forked`, one process per test. Deterministic now.

### Real numbers

| | shared interpreter | isolated (`--forked`) |
|---|---|---|
| failures | 38–43, order-dependent | **11, stable** |
| passed | — | 855 |
| skipped | 114 | 114 |

**32 of the 43 were pollution, not defects.** Only 11 are real, and they are now quarantined in `tests/known_failures.txt`:

- **1** missing optional dep (`qcelemental`) — one-line `importorskip`
- **6** assertions on source text that has since been rewritten (README, `external/CMakeLists.txt`, the int2 and scf Fortran, `oqpdata.py`)
- **1** genuine policy violation: `test_no_pyscf_dependency_in_production_runtime` forbids pyscf in the production runtime, and the FCIDUMP export path now imports it
- **1** drifted legacy-example reference
- **2** **MO symmetry labels are not invariant under rigid rotation of the input** — `b1` and `b2` swap: `Counter({'a1': 4, 'b1': 2, 'b2': 1}) != Counter({'a1': 4, 'b2': 2, 'b1': 1})`. This looks like a real bug and is the one most worth chasing.

### Final CI state

```
855 passed, 114 skipped, 11 xfailed, 55 subtests passed in 541.15s (9:01)
```

All checks green. The 855 passing tests are enforced from now on; the 11 are listed with reasons and the file is meant to shrink to nothing. A quarantined test that starts passing reports as XPASS rather than failing the run, so check the XPASS list when pruning.

Cost: **9 minutes** on one CI leg. My earlier "27 seconds" was the artifact, not a real measurement.

PR #296, opened 2026-07-28
Integrate AFQMC build and standardized input

karmachoi:
## Summary

- auto-detect an AFQMC checkout at `external/afqmc`, a sibling `afqmc`, or an explicit CMake path
- build and package the native OpenMP AFQMC executable in the same wheel as OpenQP
- add the `openqp-afqmc` installed command with a helpful fallback for builds without the private source
- add the schema-backed `[afqmc]` section, concise `afqmc(...)` aliases, and compact Python API exposure
- register the complex QR/LU LAPACK symbols required by AFQMC in the Accelerate ILP64 alias list
- add a short CH2 ROHF/STO-3G installation example and parser/API coverage

## Why

Users should be able to clone OpenQP and the private AFQMC repository, run one `pip install .`, and execute the full SCF-to-AFQMC workflow from a standardized `.oqp` input. Public OpenQP builds without private AFQMC access must continue to configure and install normally, so the native integration is enabled automatically only when the source checkout is present.

## Dependency and documentation

- AFQMC native runner and adapter: https://github.com/Open-Quantum-Platform/AFQMC/pull/1
- Required manual companion: https://github.com/Open-Quantum-Platform/openqp-docs/pull/21

## Validation

- OpenQP parser, schema, and Python API tests: 193 passed, 1 expected failure
- AFQMC adapter tests: 12 passed
- BLAS/LAPACK wrapper audit: all 169 symbols registered
- `mkdocs build --strict`: passed in the companion repository
- plain `pip install .` with a sibling AFQMC checkout: built and installed the combined wheel
- installed `openqp-afqmc` command: completed the six-step CH2/STO-3G OpenMP calculation
- `git diff --check`: clean in all three repositories


mohsenkor:
## Review (Claude)

Reviewed the full diff, reproduced the stated validation locally, and probed edge cases the added tests don't cover. **The claims in the PR description check out. I found no correctness defect.**

### Claims verified independently

| Claim | Result |
|---|---|
| "OpenQP parser, schema, and Python API tests: 193 passed, 1 expected failure" | ✅ Reproduced exactly — `193 passed, 1 xfailed` (Python 3.11, local run of `tests/test_oqp_input.py`, `tests/test_oqp_input_schema_manifest.py`, `tests/test_openqp_api.py`) |
| "BLAS/LAPACK wrapper audit: all 169 symbols registered" | ✅ `OQP_ACCELERATE_ILP64_SYMS` goes 165 → 169, exactly the four complex QR/LU routines (`zgeqrf`, `zgetrf`, `zgetri`, `zungqr`). CI rule-1 gate and both macOS Accelerate legs pass. |
| Public builds unaffected without an AFQMC checkout | ✅ All CI legs green |

Base is fresh — `main` is **0 commits ahead** of the merge base, so no stale-base effects on CI.

CI is in fact stronger evidence than the description claims: the full suite ran at **915 passed / 101 skipped / 11 xfailed**, alongside **245/245** numerical example regressions.

### The new tests actually execute

Confirmed both new tests run rather than silently skipping:

- `test_afqmc_section_call_and_concise_aliases_lower_to_legacy` — PASSED
- `TestOpenQPNativeAPI::test_settings_proxy_exposes_afqmc_section` — PASSED

### Edge cases I probed beyond the added tests

- **Alias/canonical collision** rejected in *both* orders — `afqmc(output=a,output_dir=b)` and `afqmc(output_dir=b,output=a)` both raise `Duplicate AFQMC option: output_dir`; same for `walkers`/`nwalkers`. ✅
- **Unknown keys** rejected: `afqmc(bogus_key=7)` → `Unknown option 'afqmc.bogus_key'`. ✅
- **Shipped example round-trips**: `examples/AFQMC/CH2_ROHF_STO3G_AFQMC.oqp` parse → `render_canonical_oqp` → reparse is byte-identical in the lowered legacy dict. ✅
- `CallSpec(name, args, kwargs, explicit)` field order matches the reconstruction in `_normalize_default_section_call`; the AFQMC early return only bypasses the DFTB-specific branch, which would not have applied anyway. ✅
- `{'type': string}` at `oqpdata.py:96` is **not** a typo — `string` is a real converter defined at `oqpdata.py:38`, and it is the correct choice for an enum-like key (it lowercases), consistent with `input.basis`, `mp2.variant`, etc. ✅

### `source/tdhf_mrsf_lib.F90` — safe, but out of scope and untested

This is an MRSF response-kernel change with no connection to AFQMC. I read every guarded block:

- Each `if (nbf > nocca)` / `if (nbf > noca)` wraps **only** empty-virtual-space work (`bo2v`, `bo1v`, `o21v`, and the `wrk(...,noca+1:nbf)` projections).
- The scratch array `tmp` is re-zeroed (`beta = 0.0`) by every subsequent block, so skipping leaves nothing stale behind.
- For the normal `nbf > nocca` case every guard is true, so the arithmetic is **bit-for-bit unchanged**. The 245/245 example regressions corroborate this.
- The underlying issue is real: `vb(:,nocca+1)` is an out-of-range section when `nocca == nbf`, not merely a `K=0` GEMM.
- The analogous site at line ~1055 is already covered by the pre-existing `if (nocca > 2)` guard.

**However:** no test exercises `nbf == nocca`. The shipped CH2/STO-3G example (nbf=7, nocca=5) never enters these routines, and the existing `mrsf_h2` inputs use aug-cc-pVDZ. A scientific-kernel change ideally lands with a zero-virtual-space regression case (e.g. H2/STO-3G triplet), or in its own PR.

### Non-blocking observations

1. **Silent file collision.** `install(FILES ... DESTINATION . RENAME afqmc.py)` plus `wheel.packages = ["pyoqp/oqp"]` both target `oqp/afqmc.py`. I verified empirically (scikit-build-core 1.0.3) that the CMake-installed adapter wins, so this works as designed — see the cross-check comment below. But it depends on install ordering that isn't a documented guarantee. Installing the adapter under a non-colliding name (e.g. `oqp._afqmc_adapter`) with `oqp.afqmc` as a stable dispatcher would remove the fragility.
2. **Sibling auto-detection is non-hermetic.** `ENABLE_AFQMC` defaults ON from `${CMAKE_SOURCE_DIR}/../afqmc/CMakeLists.txt` — a directory *outside* the source tree. An unrelated or stale `../afqmc` silently changes what gets built. This is deliberate per the description; flagging the trade-off, not asking for a change.
3. **Stale CI comment.** This PR removes the `Failed to import mpi4py` print from `pyoqp/oqp/utils/mpi_utils.py`, which `.github/workflows/CI.yml:195` explicitly documents a `tail -n1` workaround for. Harmless — the workaround still works — but the comment now describes behavior that no longer exists.

### Verdict

No blocking defect found. Merge-readiness is gated on process only: a required approving review and the Copilot review ruleset (the latter must be requested from the Reviewers sidebar; it cannot be triggered from the CLI). Linear history is required, so squash or rebase.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code) — cross-checked against an independent Codex review (see following comments).

mohsenkor:
## Independent review (Codex)

Second reviewer, run independently against the same diff with no access to the other reviewer's conclusions. Reported verbatim, including one finding it initially raised as a merge blocker and subsequently withdrew.

### Findings

**1. `{'type': string}` in the new `afqmc` schema — no defect.**
`string` is defined at `pyoqp/oqp/molecule/oqpdata.py:38` as `def string(strng): return strng.lower()`. Existing enum-like fields use the same converter (`input.basis:149`, `mp2.variant:174`); case-preserving paths use built-in `str`. Codex executed the module with native dependencies stubbed and confirmed `"Mean_Field"` → `"mean_field"`. No import-time `NameError`; `string` is the correct symbol for `trial`.

**2. `source/tdhf_mrsf_lib.F90` guards — numerically safe, insufficiently tested.**
The guards exclude only empty virtual-space work:
- `:666` wraps `bo2v` and `bo1v`, both explicitly OV blocks
- the `o21v` guard covers only its virtual block
- `:875` requires both a closed core and a virtual space for the CV contribution
- `:1315` and `:1351` guard only the final projection into columns `noca+1:nbf`

The general AO→MO contribution at `:1197` and the physically required OO corrections at `:1369` remain **outside** the guards. Skipped density components are zero-initialised by the callers (`tdhf_mrsf_energy.F90:540`). For normal `nbf > nocca/noca` every guard is true, so arithmetic is unchanged; for an empty virtual space the corresponding sums are mathematically empty.

One code comment is imprecise: the immediate problem in the first guarded GEMMs is the out-of-range `nocca+1` slice with `K=0`, not a zero leading dimension. The guard itself is still correct.

Only indirectly AFQMC-related. The committed CH2 mean-field example does not enter these MRSF routines, and the existing `tests/mrsf_h2` inputs use aug-cc-pVDZ with a runner that is not pytest-collected. A scientific-kernel change warrants a zero-virtual-space regression.

**3. Wheel packaging — ⚠️ RAISED AS A BLOCKER, THEN WITHDRAWN AS INCORRECT.**
Codex initially concluded that `wheel.packages = ["pyoqp/oqp"]` copies the checked-in stub over the CMake-installed adapter at `oqp/afqmc.py`, so `openqp-afqmc` would always print the fallback and exit 2 — and that the combined-wheel claim in `README.md:140` was therefore false.

This was tested empirically and refuted (see the cross-check comment). On being shown the counter-evidence, Codex verified it and responded:

> I **withdraw finding #3. It was incorrect.**
>
> The reproduction wheel contains `REAL-ADAPTER`. In scikit-build-core 1.0.3, `_build_wheel_impl_impl` runs `install_wheel()` first, then processes `wheel.packages` (wheel.py, lines 530–580). Crucially, `packages_to_file_mapping()` omits a package source when the target already exists: `not target_path.is_file()` (`_pathutil.py`, lines 104–116). Therefore CMake installs the real adapter first, and the stub is never copied over it. My ordering observation was right; my overwrite conclusion was wrong.

Worth recording: skipping a package file when the CMake install already produced that path is **explicit, intentional behaviour** in scikit-build-core, not an ordering accident. The design in this PR is sound.

**4. Sibling auto-detection is non-hermetic — non-blocking design trade-off.**
`CMakeLists.txt:78` changes `ENABLE_AFQMC` based on directories outside the source tree. Any stale or unrelated `../afqmc/CMakeLists.txt` causes its CMake code to execute and changes the wheel contents. It is also asymmetric across distribution forms: `external/*` is gitignored (`.gitignore:21`), there is no `sdist.include`/`sdist.force-include`, and scikit-build-core uses `.gitignore` for sdist exclusion by default — so a wheel built from a checkout containing `external/afqmc` auto-enables AFQMC, while a wheel rebuilt from that checkout's sdist cannot see it. Deliberate and explicitly overridable, so non-blocking.

**5. `CallSpec` construction and normalization — no defect.**
Field order `name, args, kwargs, explicit` (`oqp_input.py:58`) matches the reconstruction at `:168`. The AFQMC early return skips only the DFTB-specific removal of `backend=native` and empty `parameter_path`; no common normalization is bypassed, and schema validation still runs at `:1736`. None of `output`, `walkers`, `steps`, `dt`, `threads`, `chol` collides with a real schema key, and alias+canonical duplicates are correctly rejected. All three behaviours verified directly.

**6. Miscellaneous — no breaks.**
Removing the unconditional mpi4py warning (`mpi_utils.py:26`) is unrelated but benign; mpi4py is optional. The UTF-8 change (`oqpdata.py:825`) is coherent with the UTF-8 decoding at `:789` — unrelated and lacking a dedicated regression test, but no break found. The Fortran file still ends with byte `0a`; only an extra blank line was removed.

### Repository PR rules
- Rule 1 — pass. New complex routines registered; BLAS audit reports all 169 symbols covered.
- Rule 2 — pass. Optional-backend example present under `examples/AFQMC`.
- Rule 3 — pass. Settings proxy tested at `test_openqp_api.py:978`.
- Rule 4 — pass. Companion docs PR linked (openqp-docs#21).

### Final verdict

> **APPROVE / merge. No blocking correctness findings remain.**

Non-blocking follow-ups Codex recommends: add a zero-virtual-space (H₂/STO-3G) regression for the Fortran change or move it to a separate PR; treat the sibling auto-enable as an accepted trade-off; clean up the stale CI comment.

---
🤖 Independent review by Codex, orchestrated via [Claude Code](https://claude.com/claude-code).

mohsenkor:
## Cross-check: two independent reviews reconciled

Two reviewers examined the same diff independently — Claude and Codex — with Codex given no access to the other's conclusions. This comment reconciles them and records how the single disagreement was settled.

### Where both reviewers agreed

| Item | Claude | Codex | Status |
|---|---|---|---|
| `{'type': string}` is a real converter, not a typo | no defect | no defect | ✅ agreed |
| Fortran guards wrap only empty-virtual-space work; identical arithmetic when `nbf > nocca` | safe | safe | ✅ agreed |
| Fortran change is out of scope and has no regression test | flagged | flagged | ✅ agreed |
| `CallSpec` field order and AFQMC early return | correct | correct | ✅ agreed |
| Alias/canonical duplicates rejected; no alias collides with a schema key | verified | verified | ✅ agreed |
| Sibling `../afqmc` auto-enable is non-hermetic | flagged, non-blocking | flagged, non-blocking | ✅ agreed |
| mpi4py print removal, UTF-8 change | benign | benign | ✅ agreed |
| BLAS symbol registration (165 → 169) | verified | verified | ✅ agreed |

Both reviewers independently reproduced **193 passed, 1 xfailed**, matching the PR description exactly.

### The one disagreement — and how it was resolved

Codex raised the wheel-packaging collision as a **merge blocker**: it reasoned that `wheel.packages = ["pyoqp/oqp"]` would copy the checked-in fallback stub over the CMake-installed real adapter at `oqp/afqmc.py`, so `openqp-afqmc` would always print "installed without AFQMC" and exit 2 — making the combined-wheel claim in the README false.

Claude had flagged the same collision as a *concern* but had not resolved which file wins. Because the claim was mechanically testable, it was tested rather than argued.

**Minimal reproduction** — a standalone scikit-build-core project mirroring OpenQP's exact configuration, with no OpenQP or AFQMC source involved:

```toml
[tool.scikit-build]
wheel.install-dir = "oqp"
wheel.packages = ["pyoqp/oqp"]
```

```cmake
install(FILES "${CMAKE_SOURCE_DIR}/adapter/openqp_to_afqmc.py"
        DESTINATION . RENAME afqmc.py)
```

- `pyoqp/oqp/afqmc.py` → `MARKER = "STUB-FALLBACK"`
- `adapter/openqp_to_afqmc.py` → `MARKER = "REAL-ADAPTER"`
- Built with `pip wheel . -w dist --no-deps` (scikit-build-core 1.0.3, CPython 3.11)

**Result — the CMake-installed adapter wins:**

```
oqp/afqmc.py -> MARKER = "REAL-ADAPTER"
```

Codex then re-examined the counter-evidence and **withdrew the finding**, identifying the precise mechanism — which turns out to be a stronger result than the empirical test alone:

> `packages_to_file_mapping()` omits a package source when the target already exists: `not target_path.is_file()` (`_pathutil.py`, lines 104–116). Therefore CMake installs the real adapter first, and the stub is never copied over it. My ordering observation was right; my overwrite conclusion was wrong.

This matters for how much weight to give the result: not overwriting CMake-installed output is an **explicit guard** in scikit-build-core, not an incidental ordering artefact. The PR's design is sound, and it also means the residual fragility Claude flagged is smaller than first assessed — though installing under a non-colliding name would still make the intent self-evident to future readers.

### Net effect of the cross-check

The sole blocking finding raised across both reviews did not survive verification. Both reviewers converge on **no blocking correctness defect**.

Both also independently converge on the same non-blocking recommendation: **the `source/tdhf_mrsf_lib.F90` change is unrelated to AFQMC and ships without a zero-virtual-space (`nbf == nocca`) regression test.** It is provably safe for existing calculations — every guard is true whenever a virtual space exists — but the case it fixes is exercised by nothing in the suite. Adding an H₂/STO-3G triplet case, or splitting that hunk into its own PR, is the one substantive follow-up both reviewers agree on.

### Remaining gates

Process only, not code:
1. One approving review (`REVIEW_REQUIRED`; no reviews recorded, no unresolved threads)
2. Copilot review per the active repository ruleset — must be requested from the Reviewers sidebar; it cannot be triggered from the CLI
3. Linear history required — squash or rebase, not a merge commit

CI is green across all legs on a fresh base (`main` is 0 commits ahead of the merge base): 915 passed / 101 skipped / 11 xfailed in the Python suite, plus 245/245 numerical example regressions.

---
🤖 Two-reviewer gate orchestrated with [Claude Code](https://claude.com/claude-code).

PR #299, opened 2026-07-31
Fix UMRSF mixed exchange channels

VladimirMakhnev:
## Summary

Fix UMRSF energy evaluation for mixed alpha/beta exchange channels by correcting the 9:10 channel permutation in `tdhf_mrsf_lib.F90`.

The previous channel mapping produced incorrect UMRSF S1/S2 energies in the FC comparison against GAMESS. With the corrected mixed-channel exchange handling, OpenQP UMRSF energies agree with the GAMESS reference for the checked case.

## Changes

- Correct UMRSF mixed exchange channel handling in `source/tdhf_mrsf_lib.F90`
- Add a focused regression check for the GAMESS-compatible 9:10 channel permutation in `tests/test_umrsf_energy_regression.py`

## Validation

- Rebuilt OpenQP with the fixed UMRSF branch during local validation
- Recomputed thymine FC UMRSF energies and compared S1/S2 against GAMESS
- Ran `python tests/test_umrsf_energy_regression.py` in the OpenQP checkout


mohsenkor (approved):
Approving — CI is green across every leg.

**CI status after `0157963`:** 248/248 examples passed (0 failed, 0 errors), pytest 957 passed / 101 skipped / 11 xfailed, all gcc/arm/macOS-Accelerate legs green.

**What was blocking:** the frozen `C4H6_BHHLYP_UMRSFTDDFT_ENERGY` reference — generated in #114 against the previous 9:10 permutation — was the sole red example (max |dE| = 0.11226911 Hartree). It has been regenerated for the new permutation.

**How the regeneration was validated before it was written:**

- The branch was built locally (GCC 11.2, OpenBLAS ILP64) and the run reproduced CI's reported deviation to all eight digits (`0.11226911`) *before* the update, confirming this environment matches CI's numerics for this example.
- The SCF energy is unchanged — `-155.791338215008`, delta `8e-13` — confirming the change is confined to the TDDFT response digestion and does not perturb the reference determinant. The rewrite was scripted to abort if the SCF had moved.
- Only `td_energies` was modified; every other field in the JSON is byte-identical.

**Recorded for future readers:** the reference was regenerated from the changed code, so this example is now a self-consistency check rather than external validation. The shift is uneven — S0 −0.005 eV and S1 −0.034 eV, but S2 −1.46 eV and S3–S9 −2.6 to −3.1 eV. The GAMESS comparison cited in the PR covers thymine S1/S2, i.e. the states that barely move in this system, so the higher C4H6 roots are not independently confirmed. A GAMESS UMRSF number for C4H6 would close that gap if anyone revisits this.

**Disclosure:** I authored the reference-regeneration commit on this branch (`0157963`), so this approval is not fully independent of the branch contents. The upstream physics commit (`790d5f8`) is the author's and unmodified.

---
🤖 Reviewed and verified with [Claude Code](https://claude.com/claude-code).

PR #300, opened 2026-07-31
Fix nonconforming Fortran that aborts -fcheck=bounds builds (tagarray dims, guess maxloc mask, dftlib grow-reallocate)

karmachoi:
## Summary

Three small fixes for nonconforming Fortran that makes any `-fcheck=bounds` (Debug) build abort on a plain H2O BHHLYP MRSF energy run (`examples/MRSF-TDDFT/H2O_BHHLYP-MRSFTDDFT_ENERGY.inp`). Release builds are **not** affected numerically (details below) — the bugs only make bounds-checked debugging builds unusable.

### 1. `tagarray_driver.F90` — `dims` copy does not conform for scalar records

```fortran
if (present(dims)) dims(1:record_info%ndims) = record_info%dims
```

When a 0-dimensional (scalar) record is queried, the LHS section is zero-sized while the RHS extent differs, so bounds-checked builds abort at startup — before any calculation runs:

```
Fortran runtime error: Array bound mismatch for dimension 1 of array 'dims' (0/1)
```

`record_info%dims` is a pointer of extent `ndims` (null for scalar records). Fix: slice both sides and skip the no-data case.

### 2. `guess.F90` — `maxloc` mask does not conform to the searched section

```fortran
j = maxloc(abs(tmp(:nmo,i)), dim=1, mask=unused)
```

`unused` has length `nbf` while the searched array is the `(:nmo)` section; the `mask` argument must conform to `array` (F2018 16.9.133). Aborts with `(19/4)` on H2O 6-31G* in the Huckel guess. Fix: `mask=unused(:nmo)`.

### 3. `dftlib` — grow-reallocate uses `source=` with a non-conforming shape

`dft_molgrid.F90` (`reallocate_int`/`reallocate_real`) and `grid_storage.F90` (`extendListGrid`) grow arrays with

```fortran
allocate (nv(1:newsz), source=v)   ! size(v) /= newsz
```

With an explicit `allocate-shape-spec` the `source=` expression must have the same shape (F2018 9.7.1.2). Aborts at `dft_molgrid.F90:732` with `(452/300)` as soon as the slice arrays first grow (452/300 is exactly H2O: 3 atoms × 100 default slices, grown ×1.5 and padded to a multiple of 4). Fix: allocate to the new size, copy the surviving elements explicitly, then `move_alloc`.

## Why Release results were never affected

Inspection of gfortran's generated code for these constructs shows no out-of-bounds access occurs in unchecked builds: the nonconforming `allocate` copy loop is bounded by the source's extent (tail uninitialized but always written before read), and the nonconforming assignments iterate over the zero-sized LHS (no-op). This is purely a debuggability fix; no computed numbers change.

## Verification

Debug + `-fcheck=bounds` builds (gfortran 13.4, macOS arm64, `pip install .` defaults) of this branch with the fixes selectively reverted, running the H2O MRSF example each time:

| tree | result |
|---|---|
| `main` (no fixes) | aborts at `tagarray_driver.F90:231` `(0/1)` at startup |
| + tagarray fix | aborts at `guess.F90:250` `(19/4)` in the Huckel guess |
| + guess fix | aborts at `dft_molgrid.F90:732` `(452/300)` in the grid build |
| + dftlib fix (this branch) | **completes**; ROHF total energy and all three MRSF states match the reference JSON exactly |

## Note for reviewers testing with gfortran 14/15

gfortran 14.3/15.2 have an unrelated `-fcheck=bounds` **compiler regression**: a bogus runtime abort on perfectly valid code at `is_vector_inside_shape` (`dft_molgrid.F90`, the `vecs(:, ubound(vecs, 2))` section arguments), with a garbage index that changes run to run. It fires during any DFT grid build, before most of the code paths fixed here. To reproduce/verify this PR under `-fcheck=bounds`, use gfortran ≤ 13 (12.5 and 13.4 verified clean). We plan to report the regression to GCC separately.


mohsenkor:
@codex review

mohsenkor (approved):

PR #305, opened 2026-08-01
Add the missing second-order singles term to ROHF MP2

karmachoi:
## Summary

ROHF MP2 reports only the doubles energy. On a semicanonical ROHF reference that
is an incomplete E(2), and it was short by **1.1 to 12.1 mHa** on the systems
tested — silently, with nothing in the output to indicate a term was missing.

Semicanonicalisation diagonalises the occupied-occupied and virtual-virtual Fock
blocks, which is what makes the MP2 denominators well defined, but it does not
touch the occupied-virtual block. For ROHF that block does not vanish: the
stationarity condition fixes one *combination* of the two spins, not each spin
separately, so `f^a_ia` and `f^b_ia` are individually non-zero. The singles
contribution

```
E_S = sum_(ia,sigma) |f^sigma_ia|^2 / (e^sigma_i - e^sigma_a)
```

is therefore a genuine part of the second-order energy on such a reference.
Including it is what makes this ROHF-MBPT(2) rather than a doubles-only
truncation of it.

`E_S` is identically zero for RHF and for canonical UHF, where `f_ia` vanishes,
so nothing needed special-casing by reference type and neither of those paths
changes. The spin-component scales are defined for the doubles components, so the
singles term enters unscaled. It is printed only when non-zero, to keep a line of
zeros out of every RHF and UHF log.

## Effect

| System | `E_S` (Ha) |
| --- | --- |
| OH doublet / cc-pVDZ | −0.0027250078 |
| NH2 doublet / 6-31G | −0.0010842111 |
| O2 triplet / cc-pVDZ | −0.0121300022 |

## Verification

Three independent angles:

- **Against an outside implementation.** The same expression evaluated from
  PySCF's semicanonical Fock matrices agrees to ~1e-10 on all three systems.
- **Closed-shell invariance.** A closed-shell molecule run through ROHF now
  gives *exactly* the RHF correlation energy (−0.2031883718 both ways), with no
  singles term — as it must, since ROHF is RHF there.
- **UHF untouched.** A canonical UHF reference is bit-identical to before
  (−0.1510087705).

Three of the five new tests fail on the unpatched code; the other two are the
invariance checks, which must pass either way. Ran the full suite with and
without the patch: the only difference is the new test file.

Note that PySCF's `mp.UMP2(rohf.to_uhf())` is *not* a valid comparison here — it
uses the ROHF canonical eigenvalues as denominators rather than semicanonicalising,
so it answers a different question.


karmachoi:
@codex review

PR #308, opened 2026-08-02
AFQMC inside liboqp, with an MRSF-TDDFT root as the trial wavefunction

karmachoi:
## Summary

Phaseless AFQMC becomes a native OpenQP method. `[input] method = afqmc` dispatches to one `bind(C)` entry point that reads the converged SCF reference out of the tagarray, builds the Cholesky-factorized molecular Hamiltonian in memory, and runs the propagation. No file hand-off, no second executable, no Python in the compute path.

It replaces the previous arrangement, where a Python bridge ran OpenQP, factorized the ERIs with NumPy, wrote `HAMILTONIAN`/`FCIDUMP`/`AFQMC.KEYWORDS`, and spawned an `openqp-afqmc-native` executable built from a sibling checkout. The `ENABLE_AFQMC` / `OPENQP_AFQMC_SOURCE_DIR` build hook and the `openqp-afqmc` console script are removed with it — a plain `pip install .` now ships AFQMC with no extra checkout and no flags.

Kernels are vendored from Open-Quantum-Platform/AFQMC @ 6583aa9 under `source/afqmc/`; `afqmc_host.F90` supplies the few Gellan host symbols they use, so the vendored files stay close to upstream and can be re-synced.

## Trial wavefunctions

`[afqmc] trial` selects `mean_field`, `sf`, `mrsf` (both from a determinant file) or **`mrsf_state`** — new: OpenQP solves MRSF-TDDFT in the same job and `[afqmc] state` picks a root, expanded into its determinant basis with `tdhf_mrsf_lib`'s own `mrsfxvec`. That is what carries the spin-complete open-shell partner, so the trial is a genuine MRSF expansion rather than a relabelled SF vector. Determinants below `trial_threshold` are dropped — the main cost knob for the multideterminant overlap, force bias and local energy.

## Cholesky factorization

The bridge diagonalized the `nbf^2 x nbf^2` ERI supermatrix, O(nbf^6). This does a pivoted incomplete Cholesky in the AO basis instead, carrying only the residual diagonal and the accepted vectors and updating with DGEMV: O(nbf^2 * nchol^2), with no second `nbf^4` array. Factorizing in AO and rotating each vector afterwards is exact, not an approximation. Because the supermatrix is PSD, the largest residual diagonal at exit bounds every residual element, so the reported bound is rigorous and free.

## Validation

- **The Hamiltonian is exact.** CH2/STO-3G triplet ROHF, mean-field trial: the step-1 mixed estimator is `-38.428996645781` against a ROHF total energy of `-38.4289966458`. Cholesky residual bound 3.8e-17.
- **The MRSF trial is in the right spin sector.** Root 1 gives `<S^2> = 0.00047`, matching the `0.0000` OpenQP's own MRSF step reports for that root, which pins the determinant phase convention.
- **Every `[afqmc]` option was exercised**, including the file-based `sf`/`mrsf` trials, `oo_orbitals` (an identity `OOORBDAT` reproduces the un-rotated energy to all digits), `nlow` lower-state deflation and its collapse stop, and `force_bias_bound`.
- **Full pytest is unchanged**: 51 failures, byte-identical to the failure set of a separately built pre-change baseline on the same machine.
- A clean `pip install .` into an empty venv produces a working `openqp` that runs the shipped example.

## Performance

CH2/cc-pVDZ, nbf=24, nchol=227, 640 walkers x 100 steps, propagation only (setup ~3.2 s excluded):

| threads | before | after | ratio |
|--------:|-------:|------:|------:|
| 1  | 60.3 s | 42.0 s | 1.44x |
| 8  | 41.0 s |  5.4 s |  7.6x |
| 32 | 36.7 s |  3.3 s | 11.1x |

Total wall 39.9 s -> 6.5 s; CPU 245% -> 1400%. The 1->32 thread speedup was only 1.6x because the threaded walker loop was ~40% of the work: re-orthogonalization, the local-energy estimator and the auxiliary-field draw all ran serially. Those are now threaded or block-drawn, and the force bias and VHS build contract the Cholesky tensor in a single DGEMM pass instead of one BLAS-1 call per vector. `E(mixed)` block average is unchanged to 12 digits, and the estimator accumulates in walker order rather than via a reduction clause, so results stay independent of `OMP_NUM_THREADS`.

## Fixes in the vendored kernels

Building as a shared library exposed what `-Wl,--gc-sections` had hidden:

- `AFQMC_Greens_Functions_Z` contracted a ZGEMM over `NVar` where the operand was a pointer remap of only `NVar*NO` words, reading uninitialized stack. The extra columns hit a zero block, so it was harmless until the garbage was a NaN and `NaN*0` poisoned G.
- A call to a generic `GetInvLR` that does not exist (`GetInvLR_D` was meant).
- `walker.F90` referenced an undefined `MaxWalk`; dead code, dropped.
- `AFQMC_TFT_SQR` sized its DGEMM workspace `M*M` while writing `M*N`.
- The `<S^2>` report labelled MRSF trials as SF-CIS and vice versa.
- The Gellan global scratch (`Core_Memory::X`, a fixed 10^6-word static array) is replaced by local allocatables: it capped the basis size and would race once the walker loop is threaded.

## Input

Both forms are supported and tested. Sectioned `.inp` is explicit (`[input] method = afqmc`, plus `[tdhf] type = mrsf` for an MRSF trial). In the concise `.oqp` syntax an `afqmc(...)` call makes the job an AFQMC job, and the route then only says what is solved first:

```text
rohf/sto-3g geom="ch2.xyz" mult=3 energy afqmc(trial=mean_field)
mrsf(nstate=3)/bhhlyp/sto-3g geom="ch2.xyz" energy afqmc(trial=mrsf_state,state=1)
afqmc/sto-3g geom="ch2.xyz" mult=3 energy afqmc(walkers=256)
```

See [`examples/AFQMC`](examples/AFQMC/README.md).

## Notes for review

- The branch is cut from `agent/integrate-afqmc-build`, so it carries that branch's commit as well; the net diff against `main` is the internal AFQMC method.
- `int2e` materializes the AO ERI tensor in core (O(nbf^4)), which bounds the system size — the same limit the previous bridge had.
- AFQMC provides energies only; there are no gradients, and the `.oqp` route rejects non-`energy` drivers.


karmachoi:
Closing: this branch vendored source from the private Open-Quantum-Platform/AFQMC repository into the public openqp repository. Opened in error; the branch is being deleted.

Issue #309, opened 2026-08-02
MECP optimization issue


stecrespi:
Dear Developers,

I discovered that in the current version of OpenQP the MECP optimization does not converge to a true MECP.
I ran a few tests with our ORCA-based wrapper (https://github.com/CrespiLab/ExtOptORCA-OpenQP) vs. the native oqp optimizer and the geometric library. While our wrapper correctly converges, neither oqp nor geometric can.
They both optimize a minimum on the S0 without converging, failing to meet the energy gap criterion.
This bug could not be found from the tests because the MECP ones are capped with a limited maxit.
I append the optimization trajectories for benzene and azobenzene run with oqp and our wrapper, together with the status files for the two runs of oqp.

All the best
Stefano

[MECP_errors.zip](https://github.com/user-attachments/files/30630214/MECP_errors.zip)

karmachoi:
Dear Stefano,

Thank you for the careful report and especially for attaching the trajectories and status files — they made the diagnosis unambiguous. You are right on every point, including that the MECP tests could not have caught this.

**The cause.** `MECPOpt` had a single objective, a quadratic penalty with a fixed `gap_weight`:

```
F  = 0.5 (E_i + E_j) + w (E_j - E_i)^2
dF = 0.5 (G_i + G_j) + 2 w (E_j - E_i) (G_j - G_i)
```

Setting `dF = 0` makes the mean gradient cancel against the gap term. Those two vectors are not orthogonal, so they balance each other at a *nonzero* gap, and the residual scales as `1/w`. The optimizer is therefore converging correctly — onto a point that is not a crossing. That is exactly what your benzene file shows: `RMSD G` falls to `2e-6` while `Gap` sits at `0.14778` Hartree and never moves.

This also explains why both backends failed identically: `GeometricMECPOpt` and `OQPMECPOpt` both inherit `MECPOpt`, so `lib=oqp` and `lib=geometric` were minimizing the same objective.

We reproduced it independently on our own example, `examples/OPT/C2H4_BHHLYP-MRSFTDDFT_MECP_OQP.inp`, by raising `maxit` from 2 to 60 and changing nothing else. It starts essentially on the seam and walks away from it, reaching a 0.0333 Hartree gap after 60 steps. Raising `gap_weight` to 100 and 1000 shrinks the residual as `1/w` (0.00196, 0.00023) but never reaches the `1e-4` tolerance.

[...]

PR #311, opened 2026-08-02
Converge MECP and MECI searches with a multiplier objective

karmachoi:
Fixes the non-convergence reported in #309, and applies the same fix to MECI.

Companion documentation: [Open-Quantum-Platform/openqp-docs#24](https://github.com/Open-Quantum-Platform/openqp-docs/pull/24)

## The defect

`MECPOpt` offered a single objective, a quadratic penalty with a fixed `gap_weight`:

```
F  = 0.5 (E_i + E_j) + w (E_j - E_i)^2
dF = 0.5 (G_i + G_j) + 2 w (E_j - E_i) (G_j - G_i)
```

Setting `dF = 0` cancels the mean gradient against the gap term. The two vectors are **not orthogonal**, so they balance at a residual gap of order `1/w`, and `energy_gap` is unreachable. This is a formulation limit, not a coding slip: the optimizer converges correctly onto a point that is not a crossing.

All three backends share the objective — `GeometricMECPOpt` and `OQPMECPOpt` both inherit `MECPOpt` — which is why the reporter saw the native and geomeTRIC paths fail identically.

**MECI has the same weakness in its `penalty` objective** (`pen_incre` defaults to 1.0, so sigma never grows), but it was masked: the native `auto` path escalates to BaekA. On scipy and geomeTRIC, where `auto` resolved to `penalty` with no escalation, it was fully exposed.

## Evidence

Reproduced on this repository's own MECP example, `examples/OPT/C2H4_BHHLYP-MRSFTDDFT_MECP_OQP.inp`, with `maxit` raised from 2 to 60 and nothing else changed. The run starts essentially on the seam and walks off it. Raising `gap_weight` shrinks the residual as `1/w` but never reaches tolerance:

| step | w = 1 (old default) | w = 100 | w = 1000 |
|-----:|--------------------:|--------:|---------:|
| 20 | 0.00009 | 0.00043 | 0.00027 |
| 40 | 0.01249 | 0.00130 | 0.00024 |
| 60 | **0.03331** | **0.00196** | **0.00023** |

Gap in Hartree, tolerance `1e-4`. At `w = 100` the objective gradient is converged to `5.9e-5` while the gap is still 20x the tolerance.

The benzene S0/T1 case from #309 reproduces exactly: `quad` freezes at 0.1478 Hartree (4.0 eV), matching the trajectory attached to the issue.

## What changed

`mecp_search` now selects the MECP objective, and `meci_search` gains the same one:

- **`auglag`** (new default for both) — the augmented Lagrangian

  ```
  F  = 0.5 (E_i + E_j) + lam (E_j - E_i) + 0.5 mu (E_j - E_i)^2
  dF = 0.5 (G_i + G_j) + [lam + mu (E_j - E_i)] (G_j - G_i)
  ```

  `lam` and `mu` are constants within an evaluation, so `dF` is exactly the gradient of the reported `F`. Between evaluations they are refreshed by the first-order multiplier update `lam <- -(mean gradient . dG)/|dG|^2`, `mu <- 2 gap_sigma/|dG|`. Once `lam` settles it cancels the mean-gradient component along the branching direction, leaving a gap term and a projected mean gradient that are orthogonal; orthogonal terms cannot cancel, so `dF = 0` forces `E_j - E_i = 0` and a vanishing projected gradient *separately*. For MECP the states differ in multiplicity, so the coupling vanishes and the branching space is one-dimensional; for MECI it reuses the existing branching-plane update. Neither needs a derivative coupling. `gap_sigma` (default 10, must be positive) scales the gap term; at 1.0 the converged form is the Bearpark gradient projection.
- **`ubp`** — for MECP, the same objective under its literature name.
- **`penalty`** — Levine-Martinez smooth penalty.
- **`quad`** — retained to reproduce earlier MECP runs, and now flagged by the input checker.

`meci_search=auto` now resolves to `auglag` for two states. Multistate searches still select BaekA directly and are untouched, as is an explicit `meci_search=penalty`.

## Validation

**MECP, C2H4 example** — `auglag` reaches a `1.8e-5` gap by step 15 and stays at the `1e-5` level, where `quad` reaches `0.0333` and rising.

**MECP, benzene S0/T1** — identical inputs except `mecp_search`:

| step | `quad` | `auglag` |
|-----:|-------:|---------:|
| 1 | 0.17961 | 0.17961 |
| 30 | 0.14776 | 0.00799 |
| 60 | **0.14777** | **0.00200** |

`quad` is frozen; `auglag` is 74x closer to the seam after 60 steps and still descending.

**MECI, C2H4 example** (S1/S2, `maxit` 1 to 60, criteria restored to schema defaults, tolerance `1e-4`):

| search | gap at step 60 |
|---|---:|
| `penalty` | 1.16e-2, frozen from step 20 |
| `auto`, old (penalty 20 + BaekA) | 2.21e-3 |
| `auto` / `auglag`, new | **3.3e-4**, still descending |

Splitting the budget between auglag and BaekA measured *worse* than giving auglag the whole budget (3.9e-3): BaekA restarts its sigma ladder from scratch and gives back the progress auglag had made. BaekA is therefore held back as a rescue on the recovery budget, entered only when auglag misses the criteria.

`tests/test_crossing_objectives.py` exercises the objectives on a two-state model with an analytically known seam, without the compiled backend: `auglag` reaches the exact crossing for both MECP and MECI, its gradient matches a finite difference of its own value, `quad` reproduces its predicted `-0.495/(1 + 0.18 w)` residual, the MECI `penalty` residual is pinned as the reason the default moved, and `auglag` matches the Bearpark projection at `gap_sigma = 1`.

## Review round 2

- **The CI failure was mine.** The new test module installed its import stubs with `sys.modules.setdefault` and then assigned `__path__ = []` to the result. When an earlier test module had already imported the real `oqp.utils`, `setdefault` returned the real package and the assignment emptied its `__path__`, breaking every later `oqp.utils.*` import in the session — the 39 unrelated hessian/cphf/xtb/soc/dftb failures in the `gcc-build OFF` job. Stubs are now created unconditionally and restored when the import finishes.
- **`auglag`'s value and gradient are now consistent.** `lam` and `mu` were previously recomputed from geometry-dependent gradients on every evaluation while `dF` was returned as though both were constants. They are now frozen per evaluation, and a finite-difference test pins it. This costs convergence speed on MECI and benzene relative to the inconsistent form; the numbers above are the post-fix ones.
- **`mecp_search=baeka` is withdrawn.** Every evaluation advanced the BaekA sigma state, but MECP runs through the generic optimizer without the objective rebasing `OQPBaekAOpt` performs, so quasi-Newton secants would compare objectives formed at different sigma.
- MECI `auglag` no longer divides by zero when the mean gradient is parallel to the gap direction, `gap_sigma` is validated as finite and positive, `auglag` is registered in the concise `.oqp` MECI grammar, `mecp` gains the public `algorithm` spelling, and a C2H4 MECI `auglag` example ships in both syntaxes with a committed reference.

## Notes for review

- Both crossing examples use `maxit=1` or `2` and set every criterion except `energy_gap` to 1, so the suite never tested convergence. The reporter's diagnosis of that was correct. The new unit test is the real regression guard; making the examples themselves converge would be far too slow for CI.
- `gap_sigma=1` (plain Harvey/Bearpark) oscillated on C2H4, 10 converged, 100 was not better. The default of 10 is empirical and may want tuning on other systems.
- `dump_log` now routes `auglag`, `ubp` and `hybrid` through the MECI convergence table alongside `penalty`. `ubp` and `hybrid` were silently printing no table at all before.

Refs #309


karmachoi:
@chatgpt-codex-connector re-review please.

All CI is green now. Summary of what changed since the last review:

**CI failure (mine).** The new test module installed its stubs with `sys.modules.setdefault` and then assigned `__path__ = []` to the result. When an earlier test module had already imported the real `oqp.utils`, `setdefault` returned the real package and the assignment emptied its `__path__`, breaking every later `oqp.utils.*` import in the session. That produced the 39 unrelated hessian/cphf/xtb/soc/dftb failures. Stubs are now created unconditionally and restored when the import finishes. Separately, the MECP example reference was stale — generated before the multiplier change — which is what the macOS and arm 'Run example test suite' jobs caught; both references are regenerated.

**Review items:**

| # | Finding | Resolution |
|---|---|---|
| P1 | `auglag` gradient did not match its value | `lam` and `mu` are now frozen within an evaluation and refreshed afterwards, the ordinary first-order multiplier update. A finite-difference test pins it. This costs convergence speed relative to the inconsistent form, and the PR description now carries the post-fix numbers. |
| P1 | Zero coupling-vector norm gives NaN | Guarded; the coupling direction is dropped for that step and reseeded at the next geometry. |
| P1 | MECP BaekA needs objective rebasing | `mecp_search=baeka` withdrawn. MECP runs through the generic optimizer with no equivalent of `OQPBaekAOpt`'s `rebase_previous_objective`, so this was the right call rather than a partial fix. |
| P1 | `auglag` not registered in the concise MECI grammar | Registered, and `mecp` gains the public `algorithm` spelling that `meci` already had. |
| P1 | No example for the new method | `C2H4_BHHLYP-MRSFTDDFT_MECI_AUGLAG` added in both syntaxes with a committed reference. |
| P1 | Companion openqp-docs PR not linked | [Open-Quantum-Platform/openqp-docs#24](https://github.com/Open-Quantum-Platform/openqp-docs/pull/24), also linked in the description. |
| P2 | Nonpositive `gap_sigma` accepted | Validated as finite and positive. |

One item I did not take: the MECI `auglag` gradient still projects out the coupling direction, which no scalar constraint function can reproduce, so it is the exact gradient of the reported value only up to that projection. Dropping the projection would make MECI enforce the gap alone, which is not a conical intersection condition. It is documented in the docstring.

karmachoi:
@codex review

Round 2 addressed in 7f744b604:

| Finding | Resolution |
|---|---|
| P1 Reseed a dropped MECI coupling direction | Real bug, fixed. The zero vector satisfied `len(self.y) > 0`, so `t1` stayed zero and the direction could never recover. The record is left empty and the plane reseeds at the next geometry. |
| P2 UBP selector scaling | `mecp_search=ubp` withdrawn rather than special-cased. It aliased `auglag` but picked up `gap_sigma=10` instead of the `gap_sigma=1` that makes it the published Bearpark form, so the name promised something the selector did not deliver. MECP keeps `auglag`, `penalty`, `quad`. |
| P2 Duplicate UBP convergence block | `ubp` removed from the generic MECI block; it already has its own. |
| P2 Normalize the MECP algorithm while lowering | `lower_to_legacy` now runs `_normalized_mecp_options`; verified that `mecp(S0,T0,algorithm=quad)` emits `mecp_search=quad` with no stray `algorithm` key. |
| P1 Route the MECP algorithm through the Python API | `_set_optimize_options` picks the target key from the active runtype, so `job.workflow.mecp(algorithm=...)` sets `mecp_search`. |

On **P1 'Freeze multiplier updates across optimizer steps'** — I have not changed this, and I want to be explicit about why rather than silently declining.

Freezing `lam` and `mu` across steps as well would remove the multiplier update entirely and reduce the method to the fixed-weight penalty this PR exists to replace: the whole point is that `lam` converges to the value that cancels the mean-gradient component along the branching direction. Every adaptive-multiplier or adaptive-penalty method in this codebase has the same property — `MECIOpt.penalty` ramps `sigma` per evaluation, and BaekA advances `sigma` per evaluation too. BaekA compensates inside the native engine with `rebase_previous_objective`, which is the correct general fix, but that hook lives in `OQPBaekAOpt` and `MECPOpt`/`MECIOpt` are shared with the SciPy and geomeTRIC backends.

What this PR does fix is the part that was genuinely wrong: within a single evaluation the pair `(f, df)` is now exactly consistent, which is what a line search and a secant update read at a point. Extending `rebase_previous_objective` to the native `auglag` path is a real improvement and I would rather do it as a focused follow-up on the native runner than bolt it onto this change. Happy to be overruled if you would prefer it here.

The measured cost of the consistency fix is already recorded in the description: MECI and benzene converge more slowly than the inconsistent form did.

karmachoi:
@codex review

Round 3 addressed in 4e51079e7:

| Finding | Resolution |
|---|---|
| P1 Scope `mecp_search` to the MECP driver | Both new keys are now driver-scoped: `gap_sigma` to the two crossing drivers, `mecp_search` to MECP alone. Verified that `meci(S0,S1,mecp_search=quad)` is now rejected. The route-driver manifest still owns them as `[optimize]` schema keys, and its test is updated to the new structure. |
| P1 Add an example for the MECP penalty method | `C2H4_BHHLYP-MRSFTDDFT_MECP_PENALTY` added in both syntaxes with a committed reference. The concise `mecp` driver gains the `sigma`/`alpha`/`gap` spellings `meci` already had, which that example needs. |
| P2 Validate the MECP penalty parameters | `pen_sigma`, `pen_alpha` and `pen_incre` are validated when `mecp_search=penalty`; zero is still accepted for `pen_alpha` as its documented sentinel. |
| P2 Reject a zero MECI gradient-difference norm | Guarded, matching MECP. |

On **P1 'Keep the MECI scalar consistent with the projected gradient'** — this one I am leaving, deliberately.

The coupling direction `y` is not the gradient of any scalar constraint function; it is estimated by the branching-plane update precisely because no derivative coupling is available. Removing the projection would make MECI enforce only `E_j - E_i = 0`, which is a degeneracy condition, not a conical-intersection condition — the optimizer would be free to slide along the coupling direction. So the choice is between a value/gradient pair that is exact except in one estimated direction, and a method that converges to the wrong kind of point. The existing `ubp` makes the same trade with a worse scalar (a `gap^2/|dG|^2` ratio that diverges as the gradient difference shrinks).

It is stated in the `auglag` docstring: *dF is its exact gradient except for the y projection, which no scalar constraint function can reproduce*. If OpenQP later gains analytic derivative couplings, that projection becomes a real constraint with a real multiplier and the inconsistency disappears; that is the right time to revisit it.

karmachoi:
@codex review

Round 4 addressed in 03be43648:

| Finding | Resolution |
|---|---|
| P1 Enforce the MECP energy gap with geomeTRIC | Good catch, and it is worse than the comment says: `run_optimizer` receives the energy, gradient and step tolerances but geomeTRIC has no gap criterion at all, so this affects MECI on that backend too and predates this PR. geomeTRIC cannot be given a custom criterion through its API, so the run now states plainly that the crossing is not converged, reports the gap it ended on, and points at `maxit` or `lib=oqp`. Silently reporting success was the real defect. |
| P2 Guard a branching direction outside the prior plane | Fixed. `t2` is checked before the square root; an undefined rotation reseeds from the mean gradient as on the first step. |
| P2 Validate MECP algorithm names in the concise parser | Fixed; a typo is now rejected at parse time instead of surfacing as a ValueError deep in the run. |

Two I am not taking, with reasons rather than silence:

**P1 'Advance penalty sigma only after accepted steps'.** `MECPOpt.penalty` advances `sigma` per evaluation because that is exactly what `MECIOpt.penalty` has always done — same ramp, same `pen_incre`, same place in the call. Changing one and not the other would leave two objectives with the same name behaving differently depending on the runtype. If per-evaluation ramping is wrong it is wrong for both, and it should be fixed for both in a change that is about the penalty ramp, not about MECP convergence. Note also that `penalty` is not the default on either path.

**P1 'Add an example selecting the legacy quad method'.** `quad` is not a new capability; it is the objective this PR demonstrates cannot converge, kept solely so existing inputs reproduce. Shipping an example that selects it would advertise a method the input checker warns against, and its reference would pin a trajectory toward a point that is not a crossing. The converging objectives both have examples (`MECI_AUGLAG`, `MECP_PENALTY`, and the default `MECP_OQP`). If the coverage rule should apply to deprecated-but-retained options too, I would rather add it as a run-only example with no reference — say the word and I will.

karmachoi:
@codex review

Round 5 addressed in the latest commit — both were real and both were mine:

| Finding | Resolution |
|---|---|
| P1 Reject MECI selectors on MECP routes | Correct, and the consequence you describe is the bad part: `mecp(S0,T0,meci_search=quad)` validated, lowered a key MECP never reads, and ran the schema default instead, so the input asked for one objective and silently got another. `meci_search`, `pen_delta` and `pen_jump` are now subtracted from the MECP option set; verified rejected. |
| P2 Normalize the validated MECP algorithm before lowering | Fixed. `algorithm=AUGLAG` passed validation (which lowercases) but lowering kept the original spelling, and `MECPOpt` does an exact lookup, so it died at run time. The normalized value is stored, matching the MECI branch. Verified `AUGLAG` now lowers to `auglag`. |

This is round 5; the last two rounds have been about the option plumbing I introduced rather than the objective itself, which I take as a reasonable sign the numerical core has settled. Two items from earlier rounds remain declined with reasons in the thread above (per-evaluation `sigma` ramp in `penalty`, which matches the long-standing MECI behavior and should be changed for both or neither; and an example selecting `quad`, which is the deliberately non-converging legacy option).

karmachoi:
Round 6: fixed. `mecp_search` and `meci_search` are now folded to lower case where they are read, which covers the legacy `.inp` and Python API paths the concise-parser fix did not reach.

karmachoi:
@codex review

karmachoi:
@codex review

Round 7 addressed in 016ebf2cb. You were right twice over:

**The warning did not propagate.** `optimize()` still logged "geomeTRIC Optimizer Finished" and returned a normally completed `progress`, so a log reader or any automation still saw success while the requested crossing criterion was unmet. The verdict now goes through the same `"PyOQP: Geometry Optimization Has Not Converged"` title the native optimizer uses, and a `crossing_converged` flag records it so a caller does not have to parse the log at all.

**And my remediation advice was wrong.** I told users to raise `maxit`, but geomeTRIC stopped because it *met* its own geometry criteria, not because it ran out of steps — more iterations cannot change that. The message now points at `lib=oqp`, which is the backend that actually enforces `energy_gap`.

I deliberately did not raise an exception. The native backend logs and returns when it fails to converge rather than raising, and making geomeTRIC the only backend that aborts would be a new failure mode for existing MECI workflows that have been running this way since before this PR. If you think a gap-open crossing should be fatal, that seems like a decision for both backends together rather than something to introduce here.

karmachoi:
Round 8 in 58cb12cfd.

**P2 Expose the crossing verdict through the workflow result** — correct, the flag was unreachable. `compute_geom()` creates the optimizer as a local and discards both it and `optimize()`'s return value, so nothing outside could read it. The verdict is now set on the molecule as well, which is what `OpenQP.run()` and `Runner` return.

**P1 Propagate gap-open geomeTRIC nonconvergence** — the specific evidence in this comment is from the previous revision: `optimize()` no longer logs "geomeTRIC Optimizer Finished" on a gap-open crossing, it logs "PyOQP: Geometry Optimization Has Not Converged" with the gap it ended on. What it still does not do is raise, and that is deliberate: the native optimizer also logs and returns rather than raising when it fails to converge, so making geomeTRIC the only backend that aborts would be a new failure mode for MECI workflows that predate this PR. With this commit the verdict is also machine-readable, so automation no longer has to infer success from the absence of an exception.

karmachoi:
@codex review

karmachoi:
@codex review

Round 9 addressed in the latest commit. Both were mine and both were the same mistake in opposite directions — I validated a control too broadly and another not at all:

| Finding | Resolution |
|---|---|
| P2 Validate `gap_sigma` only for auglag searches | Fixed. The check now runs against the *effective* search, with `meci_search=auto` resolved to `auglag` for two states and `baeka` for three or more. An inherited `gap_sigma=0` no longer fails a `penalty`, `quad`, `ubp`, `hybrid` or `baeka` run over a control it never reads. |
| P2 Reject invalid effective MECP penalty weights | Fixed. `penalty` and `quad` both scale their gap term by `gap_weight`, so `gap_weight=0` removes the constraint exactly as `pen_sigma=0` would. I had validated `pen_sigma`, `pen_alpha` and `pen_incre` but left the weight open. Both objectives now require a finite positive value. |

Separately, I audited the companion manual by hand rather than by review, because openqp-docs has no Codex workflow and its later commits were never re-reviewed. That turned up two pages this PR had left stale — `baeka-multistate-meci.md` still listed `penalty`/`ubp`/`hybrid` as the two-state algorithms without mentioning the new default, and `api/python-runner.md` documented the MECI public controls but not `gap_sigma` or the fact that `workflow.mecp` now takes its own `algorithm`. Both are fixed in [openqp-docs#24](https://github.com/Open-Quantum-Platform/openqp-docs/pull/24), and I verified programmatically that the value lists in the manual now match `MECI_SEARCH` and `MECP_SEARCH` in the shipped code exactly.

karmachoi:
Reversing my earlier decline on **'Advance penalty sigma only after accepted steps'** — you were right, and checking the call sites made it worse than the comment claimed.

I had assumed the native backend called `one_step` once per accepted iteration. It does not: `_OQPRunner` hands `one_step` to `OQPEngine` as `energy_gradient` and explicitly rejects geometries from inside it. geomeTRIC calls the engine per line-search point, and so does SciPy. So the ramp advanced on **rejected** steps on every backend, and the optimizer was forming secants and trust ratios from objectives built at different sigma.

I also could not fix it the way BaekA does. `OQPBaekAOpt` gets an accepted-step signal from the native engine through `rebase_previous_objective`, but `MECPOpt` is the same class the SciPy and geomeTRIC backends use, so that hook is not reachable.

So rather than keep a ramp that cannot be trusted, **the MECP penalty now holds sigma fixed**, and the checker states plainly that `pen_incre` is not applied and points at `auglag` (which converges with no ramp at all, because the multiplier does the work) or BaekA for an adaptive one. That removes the per-evaluation mutable state from the objective entirely, so the failure mode is gone by construction rather than mitigated.

My earlier argument for declining was that MECI does the same thing and the two should change together. I still think MECI's ramp should not change inside this PR — it predates this work and belongs in a change about the penalty ramp itself — but that was an argument for leaving *MECI* alone, not for keeping a broken ramp in the new MECP path. Thanks for pushing on it.

karmachoi:
@codex review

PR #312, opened 2026-08-03
Make SQP the default MECP search on the native optimizer

karmachoi:
Follow-up to #311. **Makes SQP the default MECP search on the native optimizer, and it converges the benzene case from #309 that nothing else here does.**

Companion documentation: [Open-Quantum-Platform/openqp-docs#25](https://github.com/Open-Quantum-Platform/openqp-docs/pull/25)

## Why SQP suits MECP specifically

MECP is

```
minimise 0.5 (E_i + E_j)   subject to   c(R) = E_j - E_i = 0
```

a single equality constraint whose gradient `G_j - G_i` is analytic, because the two states differ in multiplicity and the derivative coupling vanishes. That is the whole problem: no estimated directions and nothing to tune. Each iteration solves

```
| B    gc | | p   |   | -mean_g |
|         | |     | = |         |
| gc^T  0 | | lam |   |   -c    |
```

with `B` a damped BFGS approximation to the Hessian of the Lagrangian, started from the native model Hessian. The step and the multiplier come out **together**, which is the substantive difference from `auglag`: `lam` is a result rather than a formula, so there is no penalty parameter and **`gap_sigma` is unused**. A test asserts on the parsed tree that the driver never reads it.

MECI is *not* a good fit for the same treatment, which is why this is MECP-only: its coupling direction is estimated by the branching-plane update and has no constraint function to put in the KKT rows.

## Results

`maxit=110`, only `mecp_search` changed:

| system | search | steps | converged | final gap | final rmsd_grad |
|---|---|---:|---|---:|---:|
| C2H4 | `auglag` | 60 | no | -1.0e-4 | 3.5e-3 |
| C2H4 | `sqp` | **12** | **yes** | 0.0 | 7.0e-6 |
| benzene S0/T1 | `auglag` | 60 | no | 2.0e-3 | 2.5e-2 |
| benzene S0/T1 | `sqp` | **63** | **yes** | 1.0e-8 | 1.0e-5 |

Benzene S0/T1 is the geometry from #309, where the original objective froze at a 0.1478 Hartree gap. This is the first method here to satisfy *every* criterion on it rather than only approaching the seam.

## The coordinate system is the whole story

The first version of this driver worked in Cartesians. It converged C2H4 in 48 steps but was unstable on benzene: the gradient swung by orders of magnitude and the run only landed anywhere sensible because the driver keeps the best geometry. A quasi-Newton search in Cartesians starts from an identity Hessian with no physical scale and couples every internal motion of a 12-atom ring.

Reusing the native coordinate machinery fixed it, and **delocalized internals are the right set, not TRIC**:

| `coordsys` | class returned | dimension for benzene |
|---|---|---:|
| `tric` | `RedundantInternalCoordinates` | **1011** |
| `dlc` | `DelocalizedInternalCoordinates` | 30 |
| `cart` | `CartesianCoordinates` | 36 |

`build_coordinates("tric")` returns the redundant primitive list — 1011 coordinates of rank 3N-6 — which makes a dense KKT system singular. The native engine copes because it goes through a pseudo-inverse; a direct KKT solve cannot. DLC is the non-redundant combination of the same primitives and, as its own docstring notes, is well enough conditioned for a plain solve. So this driver uses DLC and says so in the log; `coordsys` configures the native optimizer, which this search replaces. The DLC-to-redundant fallback inside `build_coordinates` is detected and drops to Cartesians rather than building a singular system.

I also tried Shanno-Phua sizing of the initial Cartesian Hessian: it cut C2H4 from 48 to 25 steps but made benzene *worse*. Moving to internals made it moot.

## Default selection

`mecp_search` defaults to `auto`, not to `sqp` directly, because SQP replaces the outer optimizer and has no meaning under the SciPy or geomeTRIC backends. `auto` resolves to `sqp` when `lib=oqp` and to `auglag` otherwise, mirroring `meci_search=auto`. An explicit `mecp_search=sqp` with another backend stays an error rather than a silent substitution.

## Design notes

- **Steps are never rejected.** There is exactly one electronic-structure evaluation per iteration, and the objective is never evaluated at a geometry the optimizer then discards — the failure mode that made the `penalty` sigma ramp unusable in #311 cannot arise here.
- Safeguards: the curvature model resets to the native model Hessian when the quadratic model points the wrong way, the trust radius follows the standard ratio test with the merit recombined at a single weight, back-transformation failures shrink the step and clear the stale prediction, a singular KKT system is reported rather than guessed, and a non-converged run returns its best geometry and re-evaluates it so the reported energies match.
- An example ships in both input syntaxes with a reference. `tests/test_crossing_objectives.py` covers the KKT solve, the singular-system guard, convergence on a model with a deliberately curved seam, and that `gap_sigma` is never read.


karmachoi:
Updated: SQP is now the **default** for MECP, not an opt-in alternative. It converges both test cases where the augmented Lagrangian converges neither, so leaving it opt-in would mean shipping a default that is known to be worse.

`mecp_search` defaults to `auto` rather than to `sqp` directly, because SQP replaces the outer optimizer and therefore has no meaning under the SciPy or geomeTRIC backends. `auto` resolves to `sqp` when `lib=oqp` and to `auglag` otherwise, mirroring how `meci_search=auto` already resolves per backend. An explicit `mecp_search=sqp` with another backend stays an error rather than a silent substitution.

The C2H4 MECP example reference is regenerated because the default path now runs a different optimizer.

@codex review

karmachoi:
Review addressed in 290926930. Four of the eleven were real defects in the driver, and one of them changed the results, so the numbers in the description are updated.

**Companion documentation:** [Open-Quantum-Platform/openqp-docs#25](https://github.com/Open-Quantum-Platform/openqp-docs/pull/25) (the P1 asking for it).

| Finding | Resolution |
|---|---|
| P1 Keep merit comparisons at a single penalty weight | Real, and it mattered. `weight` grows with the multiplier, but the stored merit kept the value computed at the old weight while the prediction and the new merit used the new one, so the ratio compared two different functions. The point is now stored as components and recombined at the current weight, which is what BaekA already does when its sigma moves. |
| P1 Preserve the geometry that satisfied convergence | Real. The best-point score covers only the gap and the KKT residual, so it could prefer a geometry that never passed the step and energy tests and overwrite the converged answer. Restoration now runs only when the search did not converge. |
| P1 Re-evaluate properties after restoring a prior geometry | Real. `mol.energies`, `mol.grads` and the metrics described the last geometry visited rather than the one returned. The restored geometry is re-evaluated. |
| P1 Link the companion openqp-docs PR | openqp-docs#25, linked above and in the description. |
| P2 Reject redundant fallbacks before solving the KKT system | Real and precisely the trap this driver exists to avoid: `build_coordinates("dlc")` falls back to the redundant set, which would make the dense KKT system singular. Detected, with Cartesians used instead. |
| P2 Keep the trust floor below the configured maximum | Fixed; the floor is now derived from the configured radii. |
| P2 Recompute the prediction after halving the step | Fixed. |
| P2 Clear the prediction after abandoning a back-transform | Fixed. |
| P2 Reject recovery controls that SQP cannot honor | The checker now warns that `auto_recovery`, `recovery_maxit` and `recovery_trust` are not read, and says which selector does use them. |
| P3 Document auto as the default MECP choice | Fixed; the README lists `auto` and no longer labels `auglag` the default. |

**The merit fix changed the benzene trajectory**, so re-measured with everything applied (maxit 110, only `mecp_search` changed):

| system | search | steps | converged | final gap | final rmsd_grad |
|---|---|---:|---|---:|---:|
| C2H4 | `auglag` | 60 | no | -1.0e-4 | 3.5e-3 |
| C2H4 | `sqp` | **12** | **yes** | 0.0 | 7.0e-6 |
| benzene S0/T1 | `auglag` | 60 | no | 2.0e-3 | 2.5e-2 |
| benzene S0/T1 | `sqp` | **63** | **yes** | 1.0e-8 | 1.0e-5 |

Benzene now needs 63 steps rather than the 56 I reported before, which means the earlier run finishing inside a 60-step budget was partly luck. Both cases converge on their own criteria; C2H4 got faster.

@codex review

karmachoi:
Round 2 addressed in the latest commit. All three were correct.

| Finding | Resolution |
|---|---|
| P2 Avoid warning on schema-injected recovery defaults | Right — the schema puts all three into every configuration, so my check fired on every native MECP run. It now compares against the defaults and speaks only when the user actually set one. |
| P2 Keep restoration evaluation within the iteration budget | Right. Re-evaluating the restored geometry pushed a maxit-limited run one step past its budget and added a status row that read like another search step. It is bookkeeping, not a step, so the iteration counter is saved and restored around it. |
| P3 Document the coordinate systems SQP actually honors | Right, my wording was wrong. `coordsys=cart` *is* honoured; what cannot be honoured is `tric`, because a dense KKT system needs a non-redundant basis, so everything else maps to DLC. The README says that now. |

C2H4 re-verified after the change: 12 steps, converged, gap 0.0, rmsd_grad 7.0e-6 — unchanged, as expected since none of these touch the step computation.

@codex review

karmachoi:
Round 3 addressed in the latest commit — you were right that restoring the counter was not enough. `record()` had already appended a status row before the `finally` block ran, so a maxit-limited run showed one more step than it took. The recorder is now swapped for one that refreshes the metrics without writing a row. Verified on benzene with an optimize `maxit` of 8: the status file has exactly 8 rows.

Scope note: this PR stays **MECP-only**. I have a working MECI SQP on a separate branch — it keeps the gap as the one genuine KKT constraint and removes the estimated coupling direction from the space the step is solved in — but that is a different discussion and should not ride along here. MECI already has two objectives that converge; MECP was the one that could not.

karmachoi:
@codex review

PR #324, opened 2026-08-05
Reorientation is left in place when the symmetry reduction does not happen

karmachoi:
`reorient_for_integral_symmetry` moves the molecule into the symmetry standard frame *before* the basis exists. Staging the petite maps can still decline after that, for reasons only visible once the basis is built. Those runs were left in the moved frame with nothing to show for it.

**This is not a rotation of symmetric molecules only.** The standard frame is centred on the charge-weighted centroid, so **every** molecule is translated — including C1, where the rotation is exactly the identity and there is no symmetry to exploit at all. Measured on a CHFClBr-like geometry: `|R − I| = 0`, `|origin| = 2.8 bohr`.

## The fix was already half-written and dead

`input_coords` is captured in the reorientation loop and **never used**, so both failure exits returned `False` with the geometry already moved. They now restore it, and the caller restores it too when staging declines — re-running `_prep_guess`, since the 1e integrals were built in the moved frame and the coordinates cannot go back without them. The cost lands only on a path that was already getting no reduction.

The coordinates are carried in `meta['_reorient_input_coords']`, deliberately **outside** the `integral_symmetry` block. Every bail in `stage_integral_symmetry_maps` replaces that block wholesale, so keeping them inside silently loses them — which is exactly what happened on the first attempt: the restore never fired for any spherical basis, because `skipped_basis_mismatch` had already wiped the key. The caller pops it immediately, so it never reaches `save_data`.

## `prop` should never have been on the allow-list

Narrowed from `('energy', 'grad', 'prop', 'properties')` to `('energy', 'grad')`.

`prop` is the only admitted runtype that consumes an **externally supplied** geometry expressed in the caller's frame: `load_previous_data` writes `OQP::xyz_old` from it, and `get_basis_overlap` overlaps that against the current — by then rotated and translated — structure. The cross-geometry MO overlap, the phase/reorder alignment built on it, and NACME are then all wrong. That is #319's finite-difference failure mode one level up.

It removes petite support from `runtype=prop`, but that capability was producing wrong overlaps, so it costs nothing that was correct. (`'properties'` was never a `run_func` key, so it admitted nothing.)

## The back-transform itself is deliberately not implemented

`docs/planned/input-to-standard.md` is rewritten to scope it, including the constraint that makes it more than an inverse map:

> The geometry cannot be restored before `save_data()` / `write_molden()` without also transforming `OQP::VEC_MO_A/B`. The AO basis is atom-centred and orientation-dependent, so writing input-frame coordinates next to standard-frame MOs gives a file that is internally inconsistent and silently wrong when read back as a guess. Either transform both or neither.

The doc also drops the draft's claim that the transform alone unlocks the optimiser, IRC and NEB — those displace the geometry every step, so they need per-step re-detection and must cope with the point group *changing* between steps. It does genuinely unlock the finite-difference case, which #319 currently solves by switching the reduction off in children.

## Verification

Water displaced off-origin to (0.3, 0.4, 0.6173):

| basis | outcome | energy | geometry |
|---|---|---|---|
| 6-31G* | reduction active | −76.0107348518 | stays in the standard frame (the documented contract) |
| cc-pVDZ | staging bails | −76.0269850877 | **restored to the input frame** |

Both match their `use_integral_symmetry=false` runs exactly, dE = 0.

Tests: new `tests/test_reorientation_is_undone_on_bail.py` 8 passed; symmetry 96 passed; prop/nac/overlap/track 45 passed; scf/grad/mrsf 185 passed.

## Note

#318 makes `skipped_basis_mismatch` rare (it was firing on every spherical basis). That does not change this PR — any bail should restore — it just moves this from the common case to a genuine corner.

### DIFF
--- a/docs/planned/input-to-standard.md
+++ b/docs/planned/input-to-standard.md
@@ -1,33 +1,84 @@
-# Implement the `input_to_standard` back-transform
+# The `input_to_standard` back-transform
 
-## The gap
+Status: **scoped, not implemented.** The two bugs this draft originally bundled
+with the feature are fixed in this PR; the transform itself remains open.
 
-`input_to_standard` is written at `pyoqp/oqp/molecule/molecule.py:489` and copied
-at `:641` and `:649`, and **consumed nowhere**. A grep over `pyoqp/`, `source/`
-and `tests/` returns only those three sites.
+## What exists
 
-Nothing back-transforms a gradient, a normal mode, or a velocity out of the
-standard frame.
+`reorient_for_integral_symmetry` moves the molecule into the symmetry standard
+frame and records the map that takes user input axes to it:
 
-## Why it matters
+    r_std = (r_input - origin) @ rotation^T
+    v_input = v_std @ rotation
 
-This single missing piece is the concrete technical reason the reorientation
-runtype allow-list exists (`molecule.py:445-447`), which keeps the optimizer,
-Hessian, IRC, NEB and NAMD in the input frame and therefore **excludes them from
-every integral-symmetry benefit**.
+It is written to `symmetry_metadata['integral_symmetry']['input_to_standard']`,
+copied forward when the maps stage, and **read by nothing**. No gradient,
+normal mode, dipole or velocity is transformed back.
 
-It is the prerequisite for widening `use_integral_symmetry` past a single-point
-energy.
+## What this PR fixed instead
 
-## Also fix while here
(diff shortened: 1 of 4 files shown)


karmachoi:
Folded into #319 — same subject (symmetry state and frame outliving the job that owns them), and it shares files with that PR, so reviewing them apart meant reading the same diff twice. The commit is preserved verbatim there.

PR #342, opened 2026-08-13
Fix zero-beta initialization across native guesses

karmachoi:
## Summary

- preserve a complete beta-spin MO basis when the beta electron count is zero while keeping the beta density exactly zero
- prevent UHF HOMO/LUMO diagnostics from indexing a nonexistent beta occupied orbital
- initialize beta orbital energies consistently for Hcore and SAP guesses
- cover Hcore, Huckel, modified Huckel, MINAO, and SAP for both UHF and ROHF zero-beta references

## Root cause

The ab initio guess path zeroed the entire beta MO matrix when `nelec_B = 0`, although later QMRSF and SCF paths still require a complete virtual beta orbital basis. The UHF gap diagnostic also indexed beta orbital zero. Hcore and SAP copied the alpha orbitals to beta but did not copy the corresponding orbital energies.

## Validation

- managed macOS arm64 build with Python 3.11, GCC/GFortran 15, OpenMP, and Accelerate ILP64; no LP64 BLAS/LAPACK leaks
- all ten native guess/SCF combinations (five guesses times UHF/ROHF) pass the zero-beta regression
- existing default-Huckel and advanced-guess tests pass
- H4 quintet QMRSF-DK smoke calculation converged in 9 SCF iterations and completed native, S3R, and Haar manifolds

This draft is based directly on the current upstream `main` and contains only the five zero-beta initialization files. The managed build and QMRSF smoke validation were performed on the descendant integration branch carrying the identical patch; upstream CI provides the clean-main integration check.

### DIFF
--- a/source/guess.F90
+++ b/source/guess.F90
@@ -50,8 +50,11 @@ subroutine get_ab_initio_density(alpha_density,alpha_orbital,beta_density,beta_o
    call orb_to_dens(alpha_density,alpha_orbital,occno,na2,nbasis,nbasis)
 
    if (nb == 0) then
-     beta_density = 0
-     if (scftype == 2) beta_orbital = 0
+     ! A fully spin-polarized reference has an empty beta density, but it still
+     ! needs a complete beta orbital basis.  In particular, the Hückel driver
+     ! has already copied its projected alpha orbitals to beta for UHF/ROHF.
+     ! Preserve those orbitals and zero only the unoccupied-spin density.
+     if (present(beta_density)) beta_density = 0.0_dp
 
    else
      select case (scftype)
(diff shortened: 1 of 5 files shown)


karmachoi:
Proactive review follow-up pushed in `382b8ef`:

- guarded the ROHF HOMO/LUMO diagnostic when the alpha space is fully occupied, avoiding `mo_e_a(nelec_a+1)` past the array bound;
- nested the optional beta-energy `size()` check under `present(mo_e_b)` because Fortran does not guarantee short-circuit evaluation;
- changed the ten native zero-beta guess probes to run in fresh processes, matching OpenQP's non-reentrant runtime initialization.

Validation:

- clean managed macOS arm64 build of native commit `8c5ff61` with Python 3.11, GCC/GFortran 15, OpenMP, and Apple Accelerate ILP64;
- cached externals reused and post-build check reported no LP64 BLAS/LAPACK leaks;
- all Hcore/Huckel/modified-Huckel/MINAO/SAP × UHF/ROHF zero-beta probes pass in isolated processes.

The branch was updated by fast-forward only; remote head is `382b8ef00b700b335abbe262c041d8466bfccb8a`.

PR #356, opened 2026-08-19
Decide PT2 route ownership inside the route, not at each call site

karmachoi:
## Summary

`sc_nevpt2_gradient_route()` raises under `[pt2] gradient=analytic`. That is correct for a run that selected SC-NEVPT2 and wrong for one that selected CASPT2, because `[pt2] gradient` is **one key shared by both analytic PT2 derivatives** — `analytic` means "AN analytic derivative", not this one.

#350 hit exactly that: `H4_CASPT2_grad` and `H4_XMS-CASPT2_grad` aborted with an SC-NEVPT2 refusal, in two separate places (the gradient dispatch **and** the energy pass, which runs a phase earlier). It was fixed by gating both call sites on `is_sc_nevpt2_configuration()`.

That works, and it leaves a trap: a third caller added without the gate reintroduces the fault, and nothing but a comment says so. This moves the decision into the function.

## The distinction that was missing

The route now returns three verdicts instead of two:

| verdict | meaning | raises under `analytic`? |
| --- | --- | --- |
| `"analytic"` | this route computes it | — |
| `"numerical"` | this route **owns** the calculation (`h0=dyall`, `contraction=strong`) and declines it, with a reason | yes |
| `OTHER_ROUTE` | the calculation belongs to a different analytic PT2 derivative | **never** |

`OTHER_ROUTE` never raises whatever `[pt2] gradient` says, so no caller can be aborted by a route that does not own its calculation — gated or not. The gates come out of both call sites, and the gradient dispatch stays silent on `OTHER_ROUTE` rather than putting an SC-NEVPT2 message in the log of every analytic CASPT2 run.

## Tests

Two of #349's scope tests asserted the old contract. They are corrected, not deleted:

- `contraction=none` and `h0=fock` move out of `test_out_of_scope_calculations_are_refused_not_approximated` into a new `test_another_routes_calculation_is_declined_not_refused`, which pins `OTHER_ROUTE` **including under `gradient=analytic`** — the case that aborted CASPT2 runs.
- The remaining refusal cases (`reference=casci`, `level_shift`) are `dyall+strong` runs this route does own, and still raise.

## Verification

No new test failures against `main` (`8fa5a1c6`): 261 in both trees on the backend-free subset, every one of them `No module named 'oqp'` — this host has no built `liboqp`, so the backend-dependent legs are carried by CI.

Behaviour is unchanged for every run that reaches a gradient today; what changes is that an ungated future caller can no longer be aborted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ for method, keyword, and build references.
 | Coupled cluster | RHF-, ROHF-, and UHF-based CCSD and CCSD(T) | Correlation energies with reference-appropriate frozen-core, Cholesky, in-core, or direct algorithms | [Guide](https://open-quantum-platform.github.io/openqp-docs/workflows/coupled-cluster/) |
 | Configuration interaction | FCI and CASCI in determinant active spaces | State energies and CI vectors at fixed molecular orbitals | [Examples](examples/WF_methods) |
 | CASSCF | State-specific CASSCF and state-averaged SA-CASSCF | Energies and analytic nuclear gradients: state specific, the weighted SA-CASSCF objective, and an individual averaged root through the coupled orbital/CI Z-vector; the legacy `method=casscf` plus `[state_average]` spelling stays on central differences | [Guide](https://open-quantum-platform.github.io/openqp-docs/workflows/casscf-gradient/) |
-| Multireference PT2 | SS/MS/XMS-CASPT2, uncontracted/strongly contracted NEVPT2, and MRMP2/MCQDPT2/XMCQDPT2 | Correlated energies and central-difference nuclear gradients for supported geometry workflows | [Examples](examples/WF_methods) |
+| Multireference PT2 | SS/MS/XMS-CASPT2, `nevpt2` / `sc-nevpt2`, and MRMP2/MCQDPT2/XMCQDPT2 | Correlated energies and central-difference nuclear gradients for supported geometry workflows | [Examples](examples/WF_methods) |
 
 #### Capabilities
 
--- a/examples/WF_methods/README.md
+++ b/examples/WF_methods/README.md
@@ -55,7 +55,15 @@ or a single one with `openqp --nompi <file>.inp` (results land in `<file>.log`).
   and `grad_ranks_per_group`). State-specific CASSCF also uses an analytic
   derivative.
 - `[state_average]` — `enabled`, `target_roots`, weights.
-- `[pt2]` — `h0` (fock=CASPT2 / dyall=NEVPT2), `contraction` (none / strong=SC-NEVPT2),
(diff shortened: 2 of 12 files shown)


karmachoi:
---

## Update: NEVPT2 now names itself

Reviewing the first commit raised the obvious question — *why are these two sharing a dispatch point at all?* The answer turned out to be an inconsistency worth fixing rather than working around.

Every other member of the PT2 family has its own `input.method`:

```
caspt2  ms-caspt2  xms-caspt2  mrmp2  mcqdpt2  xmcqdpt2
```

**NEVPT2 did not.** It had to be spelled `method=caspt2` plus `[pt2] h0=dyall`, and `contraction=strong` for the strongly contracted form. So two different theories arrived at the gradient dispatch under one name — which is exactly why routing them required inspecting options instead of reading the method.

`method=nevpt2` and `method=sc-nevpt2` now name their own zeroth-order Hamiltonian and contraction, across the legacy `[input]` layer, the canonical `.oqp` grammar, the input checker, the energy and gradient dispatch sets, and the Python API.

### Backward compatibility is the point

The option spelling is what every existing input says. It keeps working and means **exactly** the same calculation. `test_the_method_name_and_the_option_spelling_agree` pins that across all six spellings:

| spelling | resolves to |
| --- | --- |
| `method=nevpt2` | `h0=dyall`, `contraction=none` |
| `method=sc-nevpt2` | `h0=dyall`, `contraction=strong` |
| `method=caspt2` + `h0=dyall` | `h0=dyall`, `contraction=none` |
| `method=caspt2` + `h0=dyall,contraction=strong` | `h0=dyall`, `contraction=strong` |
| `method=caspt2` | `h0=fock`, `contraction=none` |

That test matters more than it looks: a divergence there would turn existing inputs into slightly different calculations while **both results stayed internally consistent**, so nothing downstream could catch it.

A method name contradicting its options (`method=nevpt2` with `h0=fock`) is **refused**, not resolved by precedence — either resolution answers a request for one theory with another, and both answers are correct numbers for their own functional.

### One trap found while wiring this

`_PT2_OWNED_KEYS` seeds every emitted `[pt2]` block with `h0=fock`. A helper that let the method imply the H0 therefore emitted a block contradicting its own method name, and the run was rejected as self-contradictory. The helper states `h0`/`contraction` explicitly instead — which is what makes the two spellings identical rather than merely equivalent. (Also removes a duplicate `"gradient"` key from that same dict literal, which silently dropped one of its own entries.)

### Not in this PR

The shipped examples still use the option spelling, deliberately — that keeps the backward-compatible path under example-suite coverage. Migrating them to the new names, and the companion `openqp-docs` update, are follow-ups.

### Verification

No new test failures against `main` on the backend-free subset. The backend legs are carried by CI.

PR #357, opened 2026-08-19
Add Intel oneAPI (ifx/icx + MKL ILP64) compiler support

karmachoi:
## Summary

OpenQP currently builds only with GCC/gfortran. This PR makes the **Intel oneAPI toolchain (ifx, icx/icpx, MKL ILP64)** a supported compiler on Linux x86_64 and adds a CI leg so it stays supported.

Every source change is a latent defect that gfortran silently tolerated and ifx either rejects at compile time or executes differently. **None of them changes gfortran results**: the same tree passes the full example suite with gfortran 13.3 / OpenBLAS ILP64.

## Validation (tpx, Ubuntu 24.04, oneAPI 2026.1.1, MKL 2026.1 ILP64)

| Build | `openqp --run_tests all` |
|---|---|
| ifx 2026.1.1 + icx + MKL ILP64 (`LINALG_LIB=auto`) | **288 passed / 0 failed / 0 errors** / 6 skipped (ddX + QM-MM examples not built) |
| gfortran 13.3 + OpenBLAS ILP64, same tree | **288 passed / 0 failed / 0 errors** / 6 skipped (identical to main) |

Before the fixes the ifx build had 59 failures (every DFT example, 0.07–0.25 Ha off) and 3 errors (MRSF segfault, two `forrtl: severe (184)` in Z-vector loops).

## What ifx exposed (all fixed here)

**Fortran source**
- `integrals/boys_lut.F90` — `fgrid` was a single 28,864-literal statement; ifx has a 41,000-token-per-statement limit. Split into 8 per-grid private slabs concatenated into the same public `fgrid` (literal sequence verified byte-identical).
- `integrals/int1.F90`, `mod_1e_primitives.F90` — removed 74 `!dir$ assume_aligned` directives. gfortran ignores them; ifx honours them and they reference variables that do not exist in scope (`sblk`/`tblk`/`xyzkin`/`xyzovl`) and split the specification part. They also assert 64-byte alignment of dummy arguments that callers never guarantee.
- `dftlib/dft.F90` (+17 `use` sites) — module `dft` → `mod_dft` (project convention). A dozen routines `use dft` *and* declare `logical :: dft`; the standard forbids the clash, ifx rejects it.
- `scf_addons.F90::calc_dft_xc_density` — **the DFT bug.** For closed-shell runs the same section `pfxc(:,1)` was passed to *both* the `fa` and `fb` dummies of `dmatd_density_blk` (argument aliasing). ifx passes each section through its own copy-in/copy-out temporary and copies the untouched `fb` temporary back last, zeroing the XC Fock matrix — E_xc was exact, only V_xc vanished, so every DFT SCF converged to the wrong density. Now a scratch array is passed for `fb` when `nfocks == 1`; `pfxc` is `contiguous, target`.
- `scf.F90`, `scf_addons.F90`, `modules/tdhf_{,sf_,mrsf_}z_vector.F90` — `int2_data = int2_xxx_t(...)` intrinsic assignment to a `class(...), allocatable` variable (auto-reallocation with a different dynamic type) → `allocate(int2_data, source=...)`. The assignment form intermittently produced corrupted component descriptors under ifx (`forrtl: severe (184): FASTMEM allocation is requested`) in the Z-vector loops: 1/16 concurrent runs failed before, 48/48 pass after.
- `scf.F90` — `pfon` is a pointer only allocated when pFON is on, yet three type-bound calls were made on it unconditionally (caught by `ifx -check pointers`). Calls guarded, pointer null-initialised.
- `mathlib.F90::solve_linear_equations` — `ipvt` allocated before the DSYSV workspace query instead of being passed unallocated.
- `integrals/int2.F90`, `dftlib/dft_gridint.F90` — `omp_set_num_threads` was called with a default (8-byte under `-i8`) integer; now `int(n, c_int)`.

**Build system**
- `external/CMakeLists.txt` — libxc, libtagarray, libecpint and the bundled NetLib LAPACK were configured as ExternalProjects with **no `CMAKE_BUILD_TYPE` at all**, so each compiled with an empty configuration (no optimisation flags) while OpenQP itself built Release. Forwarded, with the externals cache revision bumped so the new artifacts get their own namespace. (Found while porting to Windows, where the same omission additionally made MSVC mix debug and release runtimes and break the link.)
- `CMakeLists.txt` (Intel branch): `-fpscomp logicals` — ifx represents `.TRUE.` as −1, so every `logical(c_bool)` field exported through `include/oqp.h` read as 255 in Python (`got a _Bool of value 255`); `-heap-arrays 64` — ifx puts automatic arrays on the stack, which segfaults large MRSF runs inside Python's 8 MB main thread.
- `cmake/oqp_functions.cmake::findBlasLapack` — CMake's `FindBLAS` searches `$MKLROOT` as `PATHS` (after system dirs), so a distribution MKL (Ubuntu `libmkl-*` 2020) silently wins over the loaded oneAPI MKL. `$MKLROOT/lib{,/intel64}` is now prepended to `CMAKE_LIBRARY_PATH` when an Intel vendor is selected.
- `external/CMakeLists.txt` — `libtagarray` ExternalProject did not forward `CMAKE_C_COMPILER`.

**CI**
- New `intel-build` job: ubuntu-24.04, Intel apt repo (compiler-fortran + compiler-dpcpp-cpp + mkl-devel 2026.1, ~3 min install), `LINALG_LIB=auto` → MKL ILP64, asserts `liboqp.so` links `libmkl_intel_ilp64`, runs `openqp --run_tests all`. This is the Linux x86_64 source-build mandate (MKL ILP64) that the gcc legs could not exercise (CI.yml previously noted it was "validated manually on MKL machines").

## How to build with Intel

```bash
source /opt/intel/oneapi/setvars.sh
CC=icx CXX=icpx FC=ifx pip install . --config-settings=cmake.define.LINALG_LIB=auto
```

## Notes for reviewers
- `boys_lut.F90` dominates the diff (~10k lines) but is a pure re-layout of the same literals; review the first ~20 lines and the final `fgrid = reshape([fgrid_g0,...,fgrid_g7])`.
- Branch is based on `8fa5a1c6` (#350); `git merge-tree` against current `main` (#356) is clean with no overlapping files.
- Not in scope: Intel MPI in CI, ddX with Intel, Windows (this PR is the prerequisite for a Windows build).


karmachoi:
Try to perform benchmarks against gcc/gfortran combinations.

PR #360, opened 2026-08-20
Accept delocate 0.13's RPATH-free macOS wheel layout

karmachoi:
Both macOS wheel legs fail their smoke test:

```
AssertionError: .../oqp/lib/liboqp.dylib lacks package-local RPATH @loader_path
```

The wheel is fine. **delocate 0.13.0** changed what it produces: it rewrites
every `@rpath/x` load command to an explicit `@loader_path/x`, and then drops
the `LC_RPATH` it no longer needs.

```
Modifying install name in oqp/lib/liboqp.dylib
  from @rpath/libdftd4.3.dylib to @loader_path/libdftd4.3.dylib
```

That layout is package-local *by construction*, and strictly tighter than an
RPATH — an RPATH is a search list the loader could satisfy from somewhere
else, while `@loader_path/x` names exactly one file. The gate was rejecting an
improvement.

## Why this matters now

The wheel matrix runs on `release`, so nothing in day-to-day CI shows this.
Left alone, the next release publishes **no macOS wheels** and `pip install
openqp` breaks on Mac. delocate 0.13.0 is already on the runners; this arrives
on its own without anyone changing OpenQP.

## What changed

Only the post-repair check. The build-time sanitizer
(`cmake/sanitize_macos_package_rpaths.cmake.in`) is untouched and still
requires `@loader_path` on the install tree.

An empty search path is accepted **only** for a library whose dependencies are
already loader-relative. An `@rpath/` dependency with no RPATH left to resolve
it is a library that cannot load, and stays fatal — so this narrows the
accepted set rather than widening it.

One subtlety: `otool -L` prints a library's own `LC_ID_DYLIB` first, so
`liboqp.dylib` appears to depend on `@rpath/liboqp.dylib` — itself. That ID is
a name consumers link against, not an edge this file resolves, so it is not
counted. `assert_canonical_dependency_graph` already skips the self-edge for
the same reason; this matches it.

Unit tests cover both halves: the RPATH-free layout passes, a dangling
`@rpath` dependency fails.

## Alternative considered

Pinning `delocate<0.13` is one line, but it freezes the tool to avoid a change
that produced a *better* artifact, and only defers the same work to whenever
the pin is lifted.

## Verification

`tests/test_dftd4_shared_interface.py` — 16 passed.

Found while preparing #359 (Windows support); split out because it is
unrelated to Windows and blocks releases on its own.

### DIFF
--- a/.github/scripts/wheel_smoke_test.py
+++ b/.github/scripts/wheel_smoke_test.py
@@ -155,26 +155,40 @@ def assert_package_local_runtime_search_paths(
 ):
     runtime_search_paths = parse_runtime_search_paths(metadata, platform_name)
     if not runtime_search_paths:
-        # delocate >= 0.13 rewrites every ``@rpath/x`` load command to an
-        # explicit ``@loader_path/x`` and then drops the now-unused LC_RPATH.
-        # That is package-local by construction -- strictly tighter than an
-        # RPATH, which is a search list the loader could resolve elsewhere --
-        # so an empty list is correct *provided* nothing still needs a search
-        # path.  A library that kept an ``@rpath/`` dependency with no RPATH
-        # left to resolve it would not load at all, so that stays fatal.
-        #
+        # ELF has no per-dependency location: DT_NEEDED carries bare sonames,
+        # and RUNPATH/$ORIGIN is the only thing keeping resolution inside the
+        # package.  Its absence stays fatal, with no exception to inspect.
+        assert platform_name == "darwin", (
+            f"{path} lacks package-local RPATH {local_rpath}:\n{metadata}"
+        )
+        # Mach-O does: delocate >= 0.13 rewrites every ``@rpath/x`` load
+        # command to an explicit ``@loader_path/x`` and then drops the LC_RPATH
+        # it no longer needs.  That is stricter than an RPATH, which is a
+        # search list the loader could satisfy from elsewhere -- but only if
+        # EVERY load command is genuinely loader-relative, so verify that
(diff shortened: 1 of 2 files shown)


karmachoi:
Validated on real wheels, at the final strictness.

Full matrix green on `cb5c850`, the commit carrying every review fix
([karmachoi/openqp run 32359475974](https://github.com/karmachoi/openqp/actions/runs/32359475974)):

```
Build wheels (macos-arm64)     success
Build wheels (macos-x86_64)    success
Build wheels (linux-x86_64)    success
Build wheels (linux-aarch64)   success
```

Each macOS leg produced 12 wheels (cp39–cp314 × `macosx_11_0` / `macosx_15_0`)
and the smoke test — a real RHF calculation, then inspection of the repaired
binaries — passed on every interpreter.

What that demonstrates, given the review rounds this went through:

- the RPATH-free layout delocate 0.13 produces is **accepted**;
- with paths **normalized first**, so `@loader_path/../.dylibs/x` still passes
  while `@loader_path/../../../../opt/homebrew/lib/x` would not;
- with the install ID skipped **by position**, so a genuine nonlocal edge
  sharing the owner's basename is not swallowed;
- and Linux unchanged — an absent RUNPATH is still fatal there, since
  `DT_NEEDED` carries bare sonames and `$ORIGIN` is the only thing keeping
  resolution inside the package.

Without this change both macOS legs fail at the original assertion, on both
architectures.

karmachoi:
@codex review

karmachoi:
@codex review

karmachoi:
@codex review

PR #379, opened 2026-08-22
Persist MRSF state-interaction data for post-run analysis

karmachoi:
## Summary
- preserve the validated MRSF state-interaction 1-TDM in full JSON exports
- preserve the corresponding transition-dipole and AO dipole tags
- keep lean regression JSON behavior unchanged

## Why
Desktop and external analysis tools need the physical-root density data after the OpenQP process exits to generate S0->Sn NTO and attachment/detachment maps. Keeping these existing tags in the full JSON makes local and frozen bundled engines expose the same analysis contract.

## Verification
- `python3 tests/test_json_td_layout.py` using the Codex bundled Python: 4 tests passed
- `git diff --check`

No electronic-structure algorithm or saved lean regression reference is changed.


karmachoi:
@codex review

Issue #371, opened 2026-08-22
Response eigenvectors have no deterministic phase; NAC, SOC and transition-dipole signs are not reproducible


karmachoi:
Response eigenvectors carry no deterministic phase, so every sign-bearing quantity derived from them is irreproducible. The regression suite currently suppresses this per key rather than fixing it.

## Evidence

**Transition dipoles.** Rerunning `examples/UMRSF-TDDFT/C4H6_BHHLYP_UMRSFTDDFT_ENERGY` on the unmodified tree at `7bc51dbd` (GCC 13.3, OpenBLAS64 ILP64) reproduces the committed energies to 4e-10 and every `|mu_ij|` to 1e-9 — but the stored `td_trans_dipole` vectors are sign-flipped: `max |d mu| = 4.658` against the committed file. For the (0,1) element the componentwise difference is exactly twice the components.

**Effective Hamiltonian.** PR #363 records the same behaviour independently, in `pyoqp/oqp/utils/regression.py`: an off-diagonal element of the CASPT2 effective Hamiltonian on `H4_MCQDPT2` flips sign with the OpenMP thread count, "reproducible run to run in one execution context, but +6.085e-03 against the reference's own value in another (2 OpenMP threads through the `.oqp` deck)". 6.085e-03 Ha is 0.17 eV — not a small coupling.

## Why the current handling is not enough

Both are handled by marking the regression key `phase_invariant`, i.e. comparing magnitudes. That hides the symptom in the test and leaves the underlying non-determinism in the code, so:

* NAC signs, SOC matrix elements, transition dipoles and MS/XMS mixing coefficients have no reproducible sign across runs, thread counts or machines;
* the committed references contain sign patterns the code no longer produces, which is silently true today for `C4H6_BHHLYP_UMRSFTDDFT_ENERGY`;
* element-wise `| |` also discards sign information that *is* physical. Individual `H_IJ` carries the arbitrary product `s_I s_J`, but a loop product `H_IJ H_JK H_KI` carries `s_I^2 s_J^2 s_K^2 = +1` and is therefore phase-invariant and meaningful. A sign error in the coupling topology cannot be detected once every element is compared by magnitude.

## Suggested direction

[...]

PR #382, opened 2026-08-22
Give MRSF response vectors a deterministic phase (#371)

karmachoi:
Addresses the MRSF half of #371. Supersedes #381, which also touched the CI path — see below for why that was wrong.

## The defect

Nothing in the MRSF solve fixes the sign of a response vector, so every signed quantity built from one — transition dipoles, non-adiabatic couplings — carries a sign that is not a property of the calculation.

Measured against **pristine `main`**, same deck, `OMP_NUM_THREADS` 1 vs 2, raw values with signs included:

```
C4H6_BHHLYP_UMRSFTDDFT_ENERGY   td_trans_dipole   max|d| = 4.651
```

Every `|mu|` agrees to 1e-9. Only the signs move.

## The change

Pin the sign right after `trfrmb`, before any transition density or dipole is built from the vectors. The rule is **the largest-magnitude amplitude positive, with near-ties resolved by the lowest index** — an exact tie is measure-zero, but two amplitudes agreeing to round-off are not, and picking by index keeps the choice stable across platforms where `argmax` alone would not be.

Afterwards the same comparison gives **2.5e-11**.

## Scope, established by a control rather than assumed

An earlier version of this change also pinned the CI vectors. That was wrong on two counts, and running pristine `main` as a control is what showed it:

| | pristine main | main + MRSF pin only |
|---|---|---|
| `C4H6` `td_trans_dipole` | **4.651** | **2.5e-11** |
| `H4_MCQDPT2` `caspt2_effective_hamiltonian` | **1.8e-12** | 5.1e-12 |

* The CASPT2 effective Hamiltonian is **already deterministic on `main`**. `canonicalize_ci_phase` and its Fortran mirror `canonical_phase` (#366) fixed it, so the `6.085e-03` Ha sign flip #363 recorded is resolved and this change neither addresses nor affects it.
* A second, narrower convention on top of the existing one would be actively harmful. With a `1e-10` window, two coefficients differing by between `1e-10` and the established `_CI_PHASE_TIE_RTOL` of `1e-8` would have their already-canonical sign reversed — reintroducing exactly the platform dependence the convention exists to remove. Caught in review of #381.

`pyoqp/oqp/library/fci.py` is untouched here.

## References

Six references on the MRSF path stored signs the code no longer produces; regenerated. They are `phase_invariant` keys, so the suite passed either way — the regeneration makes the stored values match what the code emits, which is the state #371 is about. The CASPT2 references are untouched.

Full example suite: **291 passed, 0 failed, 6 skipped**.

## Not closing #371

The MRSF path is fixed; whether `phase_invariant` can now be narrowed, and whether any other signed quantity still lacks a convention, is not settled here.


karmachoi:
Superseded. Two things this revision got wrong, both found by measurement afterwards: the six regenerated references stored signs that are only valid for the BLAS that produced them (GCC + MKL and GCC + OpenBLAS disagree on H2O_BHHLYP_SOC state 10), and the tie window of 1e-10 was narrower than the established 1e-8. Replaced by a Fortran-only change that claims within-build reproducibility and nothing more.

PR #395, opened 2026-09-04
fix(tdhf-hess): zero the base relaxed density before accumulating into it

karmachoi:
## The defect

`trans_square` accumulates into its target, `t = t - ...`. Its target `tm` was allocated without an
initializer, while the derivative arrays allocated on the very next line carried `source=0.0_dp`.
The base relaxed excited-state density was therefore uninitialized memory plus the correct
contribution, including in the occupied-virtual blocks that `trans_square` never writes at all.
`pm` and `wm` inherited the same problem.

That base density enters the analytic relaxed density derivative through the two orbital-connection
terms `dC T C^T` and `C T dC^T`, so those two were wrong while the middle term `C dT C^T` was right.

## How it was found

Everything upstream of the density assembly was checked against an independent implementation of
the same method and matched: excitation energy 8e-08, amplitudes 2e-09, base Z vector 1e-06,
amplitude derivatives 1e-07, derivative Z right-hand side 2e-06, derivative Z vector 1.3e-06,
orbital response 6e-04, and the fixed Hessian skeleton exactly. The only quantity that disagreed
was the relaxed density derivative in the atomic-orbital basis.

The decisive test needs no second program. The code already stores the relaxed excited-state
density, and its analytic excited-state gradient is verified against a finite difference of the
energy to 1.7e-05, so that density is trustworthy. Dumping it at nineteen geometries, plus and
minus 0.002 Angstrom on each Cartesian coordinate, and central-differencing gives an
assumption-free reference for its own analytic derivative:

    symmetric part of 2 * dprel(k)  ==  central difference of the stored relaxed density

That gate failed by up to 1.45 and now passes at 5.0e-04, which is finite-difference truncation.

## Verification

H2O in STO-3G with pure TDHF for state 1, where no exchange-correlation grid is involved anywhere:

| quantity | before | after |
| --- | --- | --- |
| relaxed density derivative against the finite-difference gate, worst coordinate | 1.45 | 5.0e-04 |
| Hessian against the analytic Hessian of an independent program | 5.97e-01 | 6.8e-06 |
| Hessian against its own numerical Hessian | 5.97e-01 | 1.3e-05 |

Analytic against numerical, across functionals:

| case | before | after |
| --- | --- | --- |
| H2O / STO-3G SVWN, state 1 | 4.46 | 3.3e-03 |
| H2O / STO-3G BLYP, state 1 | 4.52 | 6.1e-03 |
| H2O / STO-3G B3LYP5, state 1 | 6.85 | 4.6e-03 |
| H2O / 6-31G B3LYP5, state 1 | 7.80e-01 | 7.0e-04 |

The residual in the density-functional cases is exchange-correlation grid noise in the numerical
reference, which differentiates an atom-centred quadrature. The grid-free Hartree-Fock case is the
clean measure, and it agrees to 1e-05. Frequencies for that case now agree with the numerical ones
to 0.01 wavenumbers, where before they were off by several hundred.

The ground-state Hessian is untouched: 2.9e-06 against numerical before and after.

## Base branch

This sits on top of `claude/rks-hessian-fock-deriv-speed-20260904`, the branch of pull request
#394, because both live on the analytic-Hessian development line and the routine this touches does
not exist on `main`. Its own change is the single commit named above, one file, eight lines.


karmachoi:
Correction on where this belongs.

Pull request #391 **already fixes this defect**. Its head commit `576be1a7`, "Fix deterministic TDHF relaxed-density assembly", zeroes `tm` immediately before the first `trans_square` call and adds `test_relaxed_density_accumulator_is_initialized_before_updates` to guard it. Nothing needs to be added there.

`576be1a7` is **not** an ancestor of this branch. The analytic-Hessian line that #394 sits on diverged before it and never received the fix, so it still carried the bug and produced Hessians that were wrong by up to 6.85 Eh/bohr^2. That is what this pull request repairs.

The fix here has been changed to use `576be1a7`'s exact form, zeroing `tm` at the same place rather than at the allocation, and carries its regression test verbatim, so the two lines agree character for character and merge without conflict. `pm` is fully assigned from `tm` and `wm` is zeroed inside `form_w_and_derivative`, so zeroing `tm` alone is sufficient.

karmachoi:
Closing: not needed.

Pull request #391 already carries this fix at its head `576be1a7`, "Fix deterministic TDHF relaxed-density assembly", which zeroes `tm` before the first `trans_square` call. I rediscovered the same defect independently on the branch behind #394, which diverged before that commit and never received it, and opened this to repair that line.

That is unnecessary. The branch behind #394 has not touched `source/modules/tdhf_hessian_density.F90` at all since the two lines split at `771e95d9`, so once #391 lands on `main` the fixed file carries through to #394 with no conflict. A merge simulation of the two lines produces zero conflicts.

The verification from this work still stands and applies to #391's fix, since it is the same one-line change. On H2O in STO-3G with pure TDHF for state 1, where no exchange-correlation grid is involved: the Hessian goes from 5.97e-01 to 6.8e-06 against an independent implementation of the same method, and from 5.97e-01 to 1.3e-05 against its own numerical Hessian, with frequencies matching to 0.01 wavenumbers. Across functionals, analytic against numerical: SVWN 4.46 to 3.3e-03, BLYP 4.52 to 6.1e-03, B3LYP5 in STO-3G 6.85 to 4.6e-03, B3LYP5 in 6-31G 7.80e-01 to 7.0e-04, the residual being grid noise in the numerical reference. The ground-state Hessian is unaffected at 2.9e-06 throughout.

The branch is left in place so the evidence stays reachable.

PR #405, opened 2026-09-17
Add analytic MRSF couplings and stable NAC-guided NAMD defaults

karmachoi:
Add analytic MRSF-TDDFT derivative couplings and use them for momentum adjustment in NAC-guided TDC surface hopping. NAMD defaults enable EDC, frustrated-hop reflection, reference continuation, bounded nuclear substepping, and conditional energy correction, allowing a minimal input:

```text
mrsf/bhhlyp/6-31g*
geom="thymine.xyz"
namd(S2,nstep=1000,dt=0.5,velocity="velocity.au")
```

Scope: this draft includes the prerequisite analytic-NAC and NAMD development history, because those implementations are absent from upstream main. It is not a four-file defaults-only PR: it retains implementation, tests, required validation tools, and examples. Rebasing onto upstream main at 1f047d562 preserves its current ILP64 build policy and integrates the overlapping XC, CPHF, derivative-integral, and TRAH changes.

Validation at `16d106144`: all applicable CI checks passed, including Linux x86/ARM with and without MPI, Intel oneAPI/MKL, both macOS Accelerate ILP64 configurations, Linux wheel, source distribution, Docker build, and BLAS/LAPACK registration. Linux x86 numerical examples: 306 passed, zero failures/errors. Python suite: 2,735 passed, 194 skipped, nine expected failures, and 238 subtests passed. [CI run](https://github.com/Open-Quantum-Platform/openqp/actions/runs/35184422799).

CI integration fixes restore XC and Z-vector procedure structure and the ROHF convergence summary, update QM/MM examples to use supported isotropic rescaling, sequence local continuation after its producer with separate output names, and support checkpoint hashing on Python 3.10. Detailed NAC diagnostics require verbose output; convergence warnings remain enabled. Regression assertions cover the current defaults and batched XC representation without relaxing numerical tolerances.

Keep this PR in draft pending companion manual updates and the separate thymine campaign validation; passing CI does not validate that trajectory ensemble.

The maximum finite gap threshold avoids rejecting restart files for a non-finite parameter. Explicit input options continue to override the defaults. Private local campaign runners are not included.

Cleanup: removed 211 development-only files, including all 23 saved NPZ arrays and six generated egg-info files. Required test inputs and imported validation programs are retained. Post-cleanup dependency checks: 393 tests and three subtests passed. Removed materials remain recoverable in Git history.

## Companion pull requests

| Repository | PR | Contents |
|---|---|---|
| [openqp-docs](https://github.com/Open-Quantum-Platform/openqp-docs) | [#60](https://github.com/Open-Quantum-Platform/openqp-docs/pull/60) | `tdc=baeck_an` and the `tlf` minor-determinant choice |
| [openqp-devkit](https://github.com/Open-Quantum-Platform/openqp-devkit) | [#1](https://github.com/Open-Quantum-Platform/openqp-devkit/pull/1) | The NAC Lagrangian derivation, its validation gates, and this PR's source-shape contracts |

Merge order is openqp-devkit#1, then openqp#410, then this PR. #410 deletes
files from this repository that exist in openqp-devkit only after devkit#1
merges, and this branch must then merge main to pick up those deletions, or
#410's repository-layout gate fails it on files it never touched.

## Which tests stayed

Fifteen tests that asserted on the *shape* of engine source -- a pivot guard
preceding a division, a subroutine carrying no state between calls, a loop nest
ordered a particular way -- moved to openqp-devkit#1. None of them ran a
calculation, and one failed on a pure hoist that changed no arithmetic.

What remains here exercises the engine: the NAMD and NAC regressions, the
Fortran selftests that compile and run an extracted routine, and the digest
identity test, which states its algebra on random matrices.

That also keeps the diff reviewable. GitHub refuses to serve a diff past 20,000
changed lines and Codex review reads that diff, so an oversized PR receives no
review and no error -- this PR sat at 24,955 lines and was answered zero times
across six review requests. It is now 16,489.


mohsenkor:
@codex review

mohsenkor:
@codex review

mohsenkor:
@codex review

karmachoi:
### Thymine ic_0349: verified cause of native TRAH stagnation

At the reported fixed geometry (Å, neutral triplet ROHF, BHHLYP/6-31G*), the ROHF Hessian in `source/scf_converger.F90::calc_h_op` omits terms from the common spatial-orbital rotation. The alpha Fock transformation needs the closed–open rotations, and the beta transformation needs the open–virtual rotations, even though these rotations leave the respective spin density unchanged. Omitting them drops coupling terms between those rotation classes. The full `[F_sigma, K]` commutator restores them; the CPHF implementation already contains this form.

| Check | Original Hessian | Corrected Hessian |
|---|---:|---:|
| Native TRAH, same 30-macroiteration limit and `conv=1e-8` | Failed after 30 | Converged in 6 |
| Final displayed incoming gradient RMS, original run | 3.0741e-6 | — |
| Independently rebuilt final-Fock gradient RMS | — | **7.9025e-9** |
| Full SCF Hessian–vector product relative error against central differences (`h=1e-4`) | 9.03e-5 | **7.47e-8** |

Corrected final energy: **−453.6636041512 Eh**. The corrected native run required no post-TRAH SOSCF/DIIS fallback and did not relax the convergence threshold. This demonstrates that the omitted Hessian terms cause the fixed-geometry convergence problem.

Mohsen’s `plain_m30` and `patched_m30` diagnostics have identical 30 printed macroiteration rows, including energies and gradients. Thus the recent noise-floor patch does not explain the onset of this fixed-geometry failure; relaxing acceptance is not a correction to the Hessian.

**Scope:** these results used base commit `16d10614469834dd2893ffba341e1d0e5b0ac026` plus an isolated, opt-in diagnostic correction (`OQP_TRAH_FULL_ROHF_FOCK=1`). The correction is not yet committed or pushed, and this is not validation of the current PR head `87aa32548df27f301e3890491abdb59b367807db`. A separate JSON-start energy/gradient inconsistency remains unresolved; the verified result above uses the Hückel/default SCF sequence. OpenTrustRegion comparisons were stopped at the author’s request, so no completed external cross-validation is claimed.

Provenance: corrected native Slurm job `5448561_0`; results archived in OpenQP.DB as `trah0349-corrected-hessian-20260918`.

karmachoi:
Implemented and pushed in b8329d65728780926a9ab2771cd8b3d1b66c699a. The shared ROHF Hessian correction now applies to both native TRAH and the OpenTrustRegion interface without an environment switch. Diagnostic numerical-derivative output was removed. Added an executable Fortran regression test of the production correction against orbital finite differences; all 8 focused TRAH tests passed. The molecular validation reported above was performed with the equivalent opt-in correction before publication; external OTR convergence remains unverified as requested.

karmachoi:
@codex review

karmachoi:
Re-ran the same ic_0349 geometry from a fresh build of the published commit b8329d65728780926a9ab2771cd8b3d1b66c699a, with no diagnostic environment switches. Neutral triplet ROHF/BHHLYP/6-31G*, native TRAH, maxit=30 and conv=1e-8 unchanged. Native TRAH converged in 6 macroiterations; independently rebuilt final-Fock gradient RMS = 7.9048e-9; final energy = -453.6636041512 Eh. No post-TRAH fallback was needed. Slurm build 5448569, calculation 5448581; calculation wall time 2m53s. This confirms the published correction reproduces the earlier diagnostic result. OTR was not re-run, as requested.

karmachoi:
@codex review

source/scf_converger.F90
```diff
@@ -3949,11 +3949,38 @@ subroutine calc_h_op(self, infos, x, x2)
...
+    n=self%nbf;na=self%nocc_a;nb=self%nocc_b
+    allocate(ka(n,n),kb(n,n),fm(n,n),tmp(n,n),a(n,n))
+    call skew_sym_k(self,x,ka,na)
+    kb=0.0_dp
+    kb(na+1:n,nb+1:na)=ka(na+1:n,nb+1:na)
+    kb(nb+1:na,na+1:n)=ka(nb+1:na,na+1:n)
+    ka(na+1:n,:)=0.0_dp;ka(:,na+1:n)=0.0_dp
+    call unpack_matrix(self%fock_ao(:,1),a)
+    tmp=matmul(a,self%mo_a);fm=matmul(transpose(self%mo_a),tmp)
+    tmp=matmul(fm,ka)-matmul(ka,fm)
```
karmachoi: [P1] Preserve symmetric energy curvature away from a stationary reference. The added full commutator differentiates the gradient in the moving orbital frame, which is not the symmetric second derivative of energy at a nonstationary reference. Executing the production correction with the existing three-orbital fixture, changing the beta CV Fock elements from -0.3 to +0.1, gives H13=0.2 and H31=-0.6; independent central mixed differences of the orbital-rotation energy give -0.2 for both. The native CG and external Davidson solvers assume symmetry, so this can invalidate their steps before SCF convergence. Include the symmetric energy-curvature correction and a nonstationary-reference regression test; the current stationary fixture does not expose this.

source/modules/namd.F90
```diff
@@ -844,8 +845,112 @@ subroutine namd_rescale_velocities(vel, ke, de)
...
+    else
+      q = -bvec + sqrt_disc
+    end if
+    if (abs(q) <= tiny(1.0_dp)) then
+      if (abs(delta_e) > tiny(1.0_dp)) return
+      gamma = 0.0_dp
+    else
+      gamma1 = q/avec
+      gamma2 = 2.0_dp*delta_e/q
+      if (abs(gamma1) <= abs(gamma2)) then
```
karmachoi: [P2] Make the equal-root choice independent of the NAC phase. For a downhill hop with zero velocity projection onto the NAC, the two quadratic roots have equal magnitude, so this comparison always selects gamma1. Executing the production routine with one unit mass, zero velocity, delta_e=-0.5, and direction=(1,0,0) gives velocity=(-1,0,0); changing only the direction sign gives velocity=(1,0,0). Thus an arbitrary electronic-state phase changes the nuclear trajectory, despite both outcomes conserving energy. Use a phase-independent tie rule and test d versus -d with v dot d=0.

karmachoi (commented):
Direct code review of the TRAH and NAMD control paths at b8329d657. Two reproducible findings below. The successful thymine convergence test remains valid, but does not cover these cases; this is not a review of the complete analytic-NAC derivation.

karmachoi:
@codex review

karmachoi:
@codex review

karmachoi:
@codex review

This is the first review request since the diff came back under GitHub's 20,000-line limit. The six earlier requests were never answered: the diff API returned `406 too_large` and Codex reads that diff, so there was nothing to fail on and no error to report. The PR is now 16,489 changed lines (93 files).

Worth a close look:

- `source/modules/tdhf_mrsf_gradient.F90` — both two-particle-density digests were rewritten in `f52dca1f` and `088f121c`. The claim is that `dc3`/`dc4`/`dd3`/`dd4` hold the same multiset of products as `dc1`/`dc2`/`dd1`/`dd2` with the two factors swapped, so `-dc1-dc2-dc3-dc4+dd1+dd2+dd3+dd4` is `2*(-dc1-dc2+dd1+dd2)`, and that the NAC `db1`/`db2` groups 5-8 repeat groups 1-4. `tests/test_mrsf_gradient_digest_identities.py` states those identities on random matrices. No timing has been measured; these are static operation counts.
- The NAC route and the NAMD defaults: EDC, frustrated-hop reflection, reference continuation, bounded nuclear substepping, conditional energy correction.

Fifteen source-shape tests moved to openqp-devkit#1 in the latest commit; the engine regressions and Fortran selftests stayed.

karmachoi:
## The analytic NAC was wrong for every spherical basis — fixed in `5cea3fc7`

Found while profiling the digest rewrite on chc3: raising the basis to cc-pVTZ aborted with `free(): invalid next size`. Chasing that turned up two defects, both the same mistake — code cloned from a working sibling with the Cartesian-to-spherical step left out, and both invisible in a Cartesian basis where the two AO index spaces coincide.

**Every analytic-NAC example and test here used 6-31G or 6-31G\*.** That is why CI was green throughout.

### What was wrong

`der_overlap_matrix_ket` (`grd1.F90`) scattered the raw Cartesian shell block at `ao_offset`, which counts actual AOs. Its sibling `der_overlap_matrix` ten lines above reduces with `reduce_der1_shell_block` first and loops `basis%naos`. For a pure f shell that wrote ten columns into a seven-wide slot and off the end of `dSket`. `der_nucattr_matrix` and `der_kinetic_matrix` were checked and already reduce correctly.

`grd2_mrsf_nac_compute_data_t_get_density` indexed its spherical densities with the **Cartesian** shell extents `grd2` hands the digest. The non-NAC `grd2_mrsf_compute_data_t` switches to Cartesian-effective, bfnrm-folded copies under `HARMONIC_ACTIVE`; the NAC type had neither the copies nor the branch.

### Measured (H2O/BHHLYP, ROHF reference, gcc 12.3 + MKL ILP64)

| case | before | after |
|---|---|---|
| cc-pVDZ S0/S1 | different on every run at 8 threads, symmetry broken | reproducible, C2v-symmetric, **equals the numerical coupling to 3e-7** (−0.03194782 vs −0.03194750) |
| cc-pVDZ S1/S2 | `NaN` | finite, symmetric |
| cc-pVTZ (f), cc-pVQZ (g) | SIGABRT | exit 0, symmetric |
| 6-31G\* (Cartesian) | 0.01345865 | 0.01345865 — unchanged |
| C4H6/6-31G\* | 0.00384946 / 0.00768018 | unchanged |

The run-to-run variation was amplification, not a race: at one thread the result was bit-stable across runs and still wrong. The 1e-13 the threaded Fock build moves the SCF by came out as an O(1) change in the coupling.

### Regression

`tests/test_mrsf_nac_spherical_basis.py` needs no external reference. H2O has a C2 axis along z that exchanges the two hydrogens, so a coupling between states of definite symmetry must reproduce itself under that operation up to one overall sign. It runs both AO conventions; the Cartesian case is the control that passed throughout. Verified to pass on the fixed build and to fail — on the spherical case only — on an unfixed one. 1.6 s.

### Also in this push

Merged `main` to pick up #410's deletions, so the rule 5 layout gate applies to this branch. Gate passes locally. 94 files, 16,716 changed lines.

@codex review — please look hardest at `grd2_mrsf_nac_build_cart` and the `usecart` branch in the NAC digest against the non-NAC pair immediately above them, and at whether any other AO-indexed consumer on this branch is missing the same step.

PR #411, opened 2026-09-19
Gate grd2 digests on spherical-basis handling (rule 6)

karmachoi:
## Why

`grd2` drives every `get_density` with **Cartesian** shell extents. A digest that indexes its densities with `basis%ao_offset` / `basis%naos`, which count the *actual* AOs, is therefore wrong whenever the basis is spherical — and only then, because in a Cartesian basis the two index spaces coincide.

That failure is silent in a Pople basis. The analytic MRSF NAC digest on the #405 branch shipped without the branch and was wrong for every spherical basis:

| | |
|---|---|
| d functions (5 vs 6) | read neighbouring AO elements: wrong values, broken molecular symmetry, and the 1e-13 run-to-run noise of the threaded SCF amplified into an O(1) change in the answer |
| f functions (7 vs 10) | wrote past the end of the array — `free(): invalid next size`, SIGABRT |

Every analytic-NAC example and test in this repository used `6-31G` or `6-31G*`, so CI was green throughout. Measured on chc3; the fix is on the #405 branch.

## What this adds

`tools/check_digest_harmonic.py` requires the `get_density` bound to each concrete `grd2_compute_data_t` extension to mention `HARMONIC_ACTIVE`.

The check is **per type, not per file**. The NAC digest was added to `tdhf_mrsf_gradient.F90`, whose *other* digest handles the spherical case correctly — a file-granularity check would have missed it entirely.

Like rules 1 and 5, CI runs the trusted base-branch copy of the script against the PR-head tree, so a PR cannot weaken its own gate.

The gate is deliberately dumb: it cannot tell a correct branch from an incorrect one, only that nobody added a digest having never considered the question. The judgment goes to the Claude review as a non-blocking rule 6, which also asks whether the PR **exercises** a spherical basis rather than only a Pople one — `cc-pVDZ` reaches the d mismatch, `cc-pVTZ` the f one.

`AGENTS.md` documents it as rule 6.

## Verification

| tree | result |
|---|---|
| this branch | PASS, 13 digests |
| upstream `main` (now carrying #405) | PASS, 13 digests |
| #405 at `58667f5b`, carrying the defect | **FAIL**, naming `grd2_mrsf_nac_compute_data_t` and the reason |
| sabotage: strip `HARMONIC_ACTIVE` from `grd2_rhf` / `grd2_uhf` / `grd2_mrsf_nac` `get_density` | **FAIL** in each case |

The gate follows the inheritance **transitively**. Matching the literal
`extends(grd2_compute_data_t)` once left `grd2_rhf_compute_data_t` and
`grd2_uhf_compute_data_t` unexamined — they extend the abstract
`grd2_hf_compute_data_t`, one hop further out — which is the very failure mode
this rule exists to catch. Before the fix, stripping `HARMONIC_ACTIVE` out of
`grd2_rhf_compute_data_t_get_density` still printed `PASS`. Coverage went from
11 digests to 13.

Rule 5 correctly rejected the new script until it was added to `TOOLS_ALLOWED` with its consumer; both gates pass now.


karmachoi:
Fixed in `5e8890c6`. The inheritance-depth gap was real, and worse than non-blocking for this gate in particular — it was blind to exactly the failure mode the rule exists to catch.

I reproduced it before changing anything: stripping `HARMONIC_ACTIVE` out of `grd2_rhf_compute_data_t_get_density` and re-running the gate still printed `PASS`. Two of the most basic digests in the tree were unexamined.

The gate now collects every type that declares a parent and takes the fixed point of "derives from `grd2_compute_data_t`", instead of matching the literal base name once. The attribute list is parsed rather than matched in a fixed order, since `abstract` can be written on either side of `extends(...)`.

Re-measured:

| tree | result |
|---|---|
| this branch | PASS, **13** digests (was 11) |
| upstream `main` | PASS, 13 |
| #405 at `58667f5b` (the original defect) | FAIL, naming `grd2_mrsf_nac_compute_data_t` |
| strip `HARMONIC_ACTIVE` from `grd2_rhf` `get_density` | **FAIL** |
| strip it from `grd2_uhf` | **FAIL** |
| strip it from `grd2_mrsf_nac` | **FAIL** |

On the other two points: the stale "10 digests" in the PR description is corrected to 13, and `python3 tools/check_digest_harmonic.py source` prints `PASS, 13 digest(s)` — you were right that it needed re-running, since #405 landing on `main` brought the NAC digest in.

AGENTS.md rule 6 now records that the check is transitive.

karmachoi:
Ran the check you couldn't execute, on the pushed head `437d8ba`:

```
$ python3 tools/check_digest_harmonic.py source
grd2 spherical-basis gate (rule 6): PASS, 13 digest(s)
```

And since you re-derived the expected set by grep rather than taking the PR's numbers, here is what the gate actually examines, so the two derivations can be compared directly:

```
grd2_cas_compute_data_t        grd2_nevpt2_compute_data_t
grd2_casao_compute_data_t      grd2_pt2_compute_data_t
grd2_fockprobe_data_t          grd2_rhf_compute_data_t
grd2_fockprobe_os_data_t       grd2_sf_compute_data_t
grd2_mp2_compute_data_t        grd2_tdhf_compute_data_t
grd2_mrsf_compute_data_t       grd2_uhf_compute_data_t
grd2_mrsf_nac_compute_data_t
```

That is a 1:1 match with your list of 13.

I also ran a behavioural sweep rather than relying on inspection: for each `get_density` binding in `source/`, strip `HARMONIC_ACTIVE` from that implementation and re-run the gate. **All 13 grd2 digests produce FAIL.** Deriving each implementation name from the binding mattered — an earlier attempt of mine guessed `grd2_cas_compute_data_t_get_density`, which does not exist (the real name is `grd2_cas_get_density`), so that case had silently tested nothing and I briefly mistook my own broken test for a gate miss.

The three `conv_*` bindings the sweep also finds belong to the `scf_conv_result` / `subconverger` hierarchies in `scf_converger.F90` — unrelated to grd2, correctly out of scope.

On the trade-off you named: agreed, and deliberate. A name collision now risks a spurious *failure* instead of a spurious *pass*. For a gate whose whole purpose is never to skip a digest silently, a false alarm someone must look at is the right side to err on.

